PR #6, opened 2022-10-05
Add cards to core

guzman-raphael:
In this PR, we:
- Add cards to core
- Utilize `mkdocs-material-insiders` (if available)
- Enable `fontawesome`
- Fix color bug in mobile navigation

Here is a preview of how the cards look rendered:

image


.docker/Dockerfile
```diff
@@ -1,6 +1,14 @@
...
 WORKDIR /main
 COPY --chown=anaconda:anaconda ./.docker/pip_requirements.txt ${PIP_REQUIREMENTS}
-RUN /entrypoint.sh echo "Dependencies installed"
+COPY --chown=anaconda:anaconda ./.docker/apk_requirements.txt ${APK_REQUIREMENTS}
+RUN \
+    /entrypoint.sh echo "Dependencies installed" && \
+    rm ${PIP_REQUIREMENTS} && \
+    rm ${APK_REQUIREMENTS}
+RUN \
+    pip install --no-cache git+https://${GITHUB_TOKEN}@github.com/guzman-raphael/mkdocs-material-insiders.git@master || \
```
kabilar, line 11: Thanks @guzman-raphael.  I am not as familiar with this step.  To clarify, should this install from your fork?

src/core.md
```diff
@@ -25,10 +69,62 @@ graph
...
+
+    [:octicons-arrow-right-24: Getting started](https://datajoint.com/docs/core/djtest-docker/)
+
+-   :fontawesome-brands-docker:{ .lg .middle } **datajoint/datajoint**
+
+    ---
+
+    Official image for managing [data pipelines](../glossary#data-pipeline).
+
+    [:octicons-arrow-right-24: Getting started](https://datajoint.com/docs/core/datajoint-python/)
```
kabilar, line 112: Should this link to the docs for that docker image?

kabilar (approved):
Thanks @guzman-raphael.  Great work!  I have added a few minor comments.

dimitri-yatsenko (approved):

.docker/Dockerfile
guzman-raphael: This is the intended so yes.

MkDocs Material has a paid subscription that allows access to exclusive features which might one day be made public but could remain exclusive. They call this version of MkDocs Material the [insiders](https://squidfunk.github.io/mkdocs-material/insiders/) version. I have subscribed on behalf of our company but did it from my user account and forked (could not signup with a GH org). I have additionally given access to my fork to certain members of the team (and CI) so they can use the features there. Here I mainly need it to use the [cards](https://squidfunk.github.io/mkdocs-material/reference/grids/#using-card-grids) feature. This strategy is within their T&C so long as we don't make that MkDocs Material Insiders code public. This strategy has the added benefit of us not paying per member of our team.

This is also a POC. If all goes well with this, I fully plan on converting this to have our GH bot user be the primary sponsor and upgrade the sponsor status to support the MkDocs Material project since we'd actively be depending on the cool things they are doing.

src/core.md
guzman-raphael: This is also intended.

The actual `Dockerfile` and repo used for the official DJ image is our `datajoint-python` repo though it is old and needs to be updated. This means we won't have a dedicated docs page for just the container image. We'll likely add a section within the DJ Python docs just for getting setup with docker. Perhaps we can just add it in the installation section [here](https://datajoint.com/docs/core/datajoint-python/latest/getting_started/).

kabilar: Thank you!

.docker/Dockerfile
kabilar: Thanks @guzman-raphael!  Sounds great.

PR #15, opened 2022-10-23
Fix partner images and 404

guzman-raphael:
- Serve in subdirectory to fix 404 issue
- Fix centered images in figures
- Add HubSpot chatbot

> :warning: Note: Upon merging, we should update our [ingress](https://github.com/datajoint-company/dj-gitops/blob/0e3378e1d05dc0ba828a6180c2176e9f6b66dff0/applications/k8s/deployments/documentation/ingress.yaml#L35) to reflect the new directory scheme from MkDocs Material
> from:
> ```nginx
> rewrite ^/[^/]+(/.*)$ /datajoint-docs$1 break;
> ```
> to:
> ```nginx
> rewrite ^/[^/]+(/.*)$ /datajoint-docs/docs$1 break;
> ```

Depends on datajoint-company/dj-gitops#60

### DIFF
--- a/mkdocs.yaml
+++ b/mkdocs.yaml
@@ -1,8 +1,9 @@
 # ---------------------- PROJECT SPECIFIC ---------------------------
 
 site_name: DataJoint Documentation
-repo_url: https://github.com/datajoint-company/datajoint-docs
+site_url: http://localhost/docs
 repo_name: datajoint-company/datajoint-docs
+repo_url: https://github.com/datajoint-company/datajoint-docs
 nav:
   - Welcome: welcome.md
   - Core: core.md
@@ -39,6 +40,7 @@ nav:
       - INCF: community/partnerships/incf.md
   - Projects: projects.md
   - Publications: publications.md
+
 # ---------------------------- STANDARD -----------------------------
 
 docs_dir: ./src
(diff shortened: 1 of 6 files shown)


dimitri-yatsenko (dismissed):

mkdocs.yaml
```diff
@@ -1,8 +1,9 @@
 # ---------------------- PROJECT SPECIFIC ---------------------------
 
 site_name: DataJoint Documentation
-repo_url: https://github.com/datajoint-company/datajoint-docs
+site_url: http://localhost/docs
```
kabilar, line 4:
```suggestion
site_url: https://www.datajoint.com/docs
```

kabilar: Perhaps this suggestion is not required based on how we deploy the site.

kabilar (changes requested):
Thanks @guzman-raphael!  One minor comment.

mkdocs.yaml
guzman-raphael: Thanks @kabilar but this was intentional. Using `localhost` should protect us from future updates to the domain without requiring cascading PR's (though subdomain is a different story). I don't expect it to affect the deployment.

kabilar: Thank you

kabilar (approved):

PR #30, opened 2022-12-02
Add researchers/labs from old pubs

CBroz1:
- [X] Review full list of pubs. For any pub without corresponding project, add lab
- [X] Pub `others` -> APA format `...`

### DIFF
--- a/src/publications.md
+++ b/src/publications.md
@@ -52,7 +52,7 @@ The following publications relied on DataJoint open-source software for data ana
 
 ## 2021 <!-- 14 -->
 
-+ Bae, J. A., Baptiste, M., Bodor, A. L., Brittain, D., Buchanan, J., Bumbarger, D. J., Castro, M. A., Celii, B., Cobos, E., Collman, F.others. (2021). <a href="https://doi.org/10.1101/2021.07.28.454025" target="_blank" style="color:#1589F0;">Functional connectomics spanning multiple areas of mouse visual cortex</a>. *bioRxiv*.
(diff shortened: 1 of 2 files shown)


src/publications.md
```diff
@@ -52,7 +52,7 @@ The following publications relied on DataJoint open-source software for data ana
 
 ## 2021 <!-- 14 -->
 
-+ Bae, J. A., Baptiste, M., Bodor, A. L., Brittain, D., Buchanan, J., Bumbarger, D. J., Castro, M. A., Celii, B., Cobos, E., Collman, F.others. (2021). <a href="https://doi.org/10.1101/2021.07.28.454025" target="_blank" style="color:#1589F0;">Functional connectomics spanning multiple areas of mouse visual cortex</a>. *bioRxiv*.
++ Bae, J. A., Baptiste, M., Bodor, A. L., Brittain, D., Buchanan, J., Bumbarger, D. J., Castro, M. A., Celii, B., Cobos, E., Collman, F., ... (2021). <a href="https://doi.org/10.1101/2021.07.28.454025" target="_blank" style="color:#1589F0;">Functional connectomics spanning multiple areas of mouse visual cortex</a>. *bioRxiv*.
```
JaerongA, line 55: why don't we use `, et al.` instead?

CBroz1: For APA citations, the proper format after the 19th author is `...` for a full citation list ([source](https://www.easybib.com/guides/citation-guides/apa-format/citing-source-with-multiple-authors-apa/)). We don't have 19 in each case, but I think we're better off following the standard

JaerongA (approved):

PR #41, opened 2023-02-28
Command to activate conda environment fixed

virenvarma007:
Originally to create and activate a conda environment for DataJoint the following commands were used:

conda create --name datajoint-workflow-<name> python=3.9  conda activate dj-workflow-<name> 

which has been changed to the following as the environment created and the one activated had a different name. Thus I propose the following change to keep it constant and throw the error 'conda environment not found'.

conda create --name datajoint-workflow-<name> python=3.9  conda activate datajoint-workflow-<name>


kabilar (approved):
Thanks @virenvarma007!

PR #68, opened 2024-04-11
Update publications.md

dimitri-yatsenko:
add Spyglass 2024


ttngu207 (approved):

src/projects/publications.md
CBroz1, line 1: We had to add `Nevers` to our codespell config

ttngu207: I'd just let it fail the spellcheck this one time and just merge.
Allowing "nevers" to pass spellcheck is a bit dangerous

PR #118, opened 2026-01-14
Clarify hash-addressed vs schema-addressed storage distinction

dimitri-yatsenko:
## Summary

Clarifies the technical distinction between hash-addressed and schema-addressed storage based on their capability to handle different object types.

## Changes

**Key distinction added:**
- **Hash-addressed storage** (<blob@>, <attach@>, <hash@>): Handles **individual/atomic objects only** - single files or serialized blobs
- **Schema-addressed storage** (<npy@>, <object@>): Can handle **complex multi-part objects** - Zarr arrays, HDF5 datasets, directory structures

### src/how-to/use-object-storage.md
- Updated OAS addressing schemes table with new "Object Type" column
- Added explicit key distinction callout
- Fixed terminology: inline → in-table storage

### src/reference/specs/type-system.md
- Enhanced <object@> description: "complex, multi-part objects (files, folders, Zarr arrays, HDF5)"
- Enhanced <hash@> description: "individual, atomic objects only"
- Clarified that hash-addressed cannot handle Zarr/HDF5

### src/llms-full.txt
- Regenerated to include updated documentation

## Technical Background

This distinction is important because:
1. Zarr arrays are directory structures with multiple chunks/metadata files
2. Hash-addressed storage stores single blobs identified by content hash
3. Schema-addressed storage uses hierarchical paths that can contain multiple files
4. Users choosing between <blob@> and <object@> need to understand this limitation

## Related

Addresses feedback about clarifying storage capabilities for complex data structures.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dimitri-yatsenko:
Consolidated into #119 - Documentation Cohesion Review: Comprehensive Improvements for DataJoint 2.0

PR #155, opened 2026-02-22
docs: document restricted diagram operations (new in 2.2)

dimitri-yatsenko:
## Summary

Document the graph-driven diagram operations introduced in DataJoint 2.2. The Diagram is an inspection and graph computation tool — all mutation (delete, drop) is executed by `Table.delete()` and `Table.drop()`.

### Changes

- **`diagram.md`**: Add operational methods section documenting `cascade()`, `restrict()`, `preview()`, `prune()` with restriction propagation rules, OR-vs-AND convergence semantics, and `part_integrity` behavior. No `delete()` or `drop()` on Diagram.
- **`whats-new-22.md`**: Add graph-driven diagram operations section covering motivation, preview-then-execute pattern (Diagram for inspection → `Table.delete()` for execution), two propagation modes, architecture, and comparison table vs error-driven cascade.
- **`delete-data.md`**: Add "Inspecting Cascade Before Deleting" section showing `delete(dry_run=True)` for quick preview and `dj.Diagram(schema).cascade().preview()` for detailed inspection. Deletion always through `Table.delete()`.
- **`data-manipulation.md`**: Version admonition noting graph-driven cascade internals in 2.2.
- **`master-part.md`**: Document `part_integrity="cascade"` upward propagation behavior.
- Cross-references between all affected files.

### Key design decision

`Diagram` has no `delete()` or `drop()` methods. It provides `cascade()` and `preview()` for understanding the blast radius; `Table.delete()` handles transactions, SQL execution, prompts, and error handling. This keeps the Diagram as a pure graph computation tool.

## Test plan

- [x] `mkdocs build` succeeds with no new warnings
- [ ] Review rendered pages for diagram.md, whats-new-22.md, delete-data.md
- [ ] Verify cross-reference links resolve correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/explanation/whats-new-22.md
+++ b/src/explanation/whats-new-22.md
@@ -234,15 +234,13 @@ This is valuable when working with unfamiliar pipelines, large datasets, or mult
 
 ### Two Propagation Modes
 
-The diagram supports two restriction propagation modes with different convergence semantics:
+The diagram supports two restriction propagation modes designed for fundamentally different tasks.
 
-**`cascade()` uses OR at convergence.** When a child table has multiple restricted ancestors, the child row is affected if *any* parent path reaches it. This is the right semantics for delete — if any reason exists to remove a row, it should be removed. `cascade()` is one-shot: it can only be called once on an unrestricted diagram.
+**`cascade()` prepares a delete.** It takes a single restricted table expression, propagates the restriction downstream through all descendants, and **trims the diagram** to the resulting subgraph — ancestors and unrelated tables are removed entirely. Convergence uses OR: a descendant row is marked for deletion if *any* ancestor path reaches it, because if any reason exists to remove a row, it should be removed. `cascade()` is one-shot and is always followed by `preview()` or `delete()`.
 
-**`restrict()` uses AND at convergence.** A child row is included only if *all* restricted ancestors match. This is the right semantics for data subsetting and export — only rows satisfying every condition are selected. `restrict()` is chainable: call it multiple times to build up conditions from different tables.
+**`restrict()` selects a data subset.** It propagates a restriction downstream but **preserves the full diagram**, allowing `restrict()` to be called again from a different seed table. This makes it possible to build up multi-condition subsets incrementally — for example, restricting by species from one table and by date from another. Convergence uses AND: a descendant row is included only if *all* restricted ancestors match, because an export should contain only rows satisfying every condition. After chaining restrictions, use `prune()` to remove empty tables and `preview()` to inspect the result.
 
-Both modes propagate **downstream only** — from the seed table to its descendants. `cascade()` goes further: it trims the returned Diagram to the cascade subgraph, removing all ancestors and unrelated tables. This means `delete()` operates on the entire trimmed diagram with no additional filtering. `restrict()` keeps the full graph intact (to support chaining from multiple seed tables) but only restricts the seed's descendants. This matches the semantics of foreign key cascades: deleting a session deletes its trials, not its subject.
-
-The two modes are mutually exclusive on the same diagram. This prevents accidental mixing of incompatible semantics.
+The two modes are mutually exclusive on the same diagram. This prevents accidental mixing of incompatible semantics — a delete diagram should never be reused for subsetting, and vice versa.
 
 ### Pruning Empty Tables
 
--- a/src/explanation/whats-new-22.md
+++ b/src/explanation/whats-new-22.md
@@ -1,6 +1,6 @@
 # What's New in DataJoint 2.2
 
-DataJoint 2.2 introduces **isolated instances**, **thread-safe mode**, and **graph-driven diagram operations** for applications that need multiple independent database connections, explicit cascade control, and operational use of the dependency graph.
+DataJoint 2.2 introduces **isolated instances** and **thread-safe mode** for applications that need multiple independent database connections, and **graph-driven diagram operations** that replace the legacy error-driven cascade with a reliable, inspectable approach for all users.
 
 > **Upgrading from 2.0 or 2.1?** No breaking changes. All existing code using `dj.config` and `dj.Schema()` continues to work. The new Instance API is purely additive.
 
@@ -207,27 +207,29 @@ DataJoint 2.2 promotes `dj.Diagram` from a visualization tool to an operational 
 
 ### From Visualization to Operations
 
-In prior versions, `dj.Diagram` existed solely for visualization — drawing the dependency graph as SVG or Mermaid output. The cascade logic inside `Table.delete()` traversed dependencies independently, with no way to inspect or control the cascade before it executed.
+In prior versions, `dj.Diagram` existed solely for visualization — drawing the dependency graph as SVG or Mermaid output. The cascade logic inside `Table.delete()` traversed dependencies independently using an error-driven approach: attempt `DELETE` on the parent, catch the foreign key integrity error, parse the error message to discover which child table is blocking, then recursively delete from that child first. This had several problems:
 
-In 2.2, `Table.delete()` and `Table.drop()` delegate internally to `dj.Diagram`. The user-facing behavior of `Table.delete()` is unchanged, but the diagram-level API is now available as a more powerful interface for complex scenarios.
+- **MySQL 8 with limited privileges** returns error 1217 (`ROW_IS_REFERENCED`) instead of 1451 (`ROW_IS_REFERENCED_2`), which provides no table name — the cascade crashes with no way to proceed.
+- **PostgreSQL** aborts the entire transaction on any error, requiring `SAVEPOINT` / `ROLLBACK TO SAVEPOINT` round-trips for each failed delete attempt.
+- **Fragile error parsing** across MySQL versions and privilege levels, where different configurations produce different error message formats.
+
+In 2.2, `Table.delete()` and `Table.drop()` use `dj.Diagram` internally to compute the dependency graph and walk it in reverse topological order — deleting leaves first, with no trial-and-error needed. The user-facing behavior of `Table.delete()` is unchanged. The Diagram's `cascade()` and `preview()` methods are available as a public inspection API for understanding cascade impact before executing.
 
 ### The Preview-Then-Execute Pattern
 
-The key benefit of the diagram-level API is the ability to build a cascade explicitly, inspect it, and then decide whether to execute:
+The key benefit of the diagram-level API is the ability to build a cascade explicitly, inspect it, and then execute via `Table.delete()`:
 
 ```python
-# Build the dependency graph
+# Build the dependency graph and inspect the cascade
 diag = dj.Diagram(schema)
-
-# Apply cascade restriction — nothing is deleted yet
 restricted = diag.cascade(Session & {'subject_id': 'M001'})
 
 # Inspect: what tables and how many rows would be affected?
 counts = restricted.preview()
 # {'`lab`.`session`': 3, '`lab`.`trial`': 45, '`lab`.`processed_data`': 45}
 
-# Execute only after reviewing the blast radius
-restricted.delete(prompt=False)
+# Execute via Table.delete() after reviewing the blast radius
+(Session & {'subject_id': 'M001'}).delete(prompt=False)
 ```
 
 This is valuable when working with unfamiliar pipelines, large datasets, or multi-schema dependencies where the cascade impact is not immediately obvious.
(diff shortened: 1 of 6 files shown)


ttngu207 (dismissed):

src/explanation/whats-new-22.md
```diff
@@ -1,6 +1,6 @@
 # What's New in DataJoint 2.2
 
-DataJoint 2.2 introduces **isolated instances** and **thread-safe mode** for applications that need multiple independent database connections—web servers, multi-tenant notebooks, parallel pipelines, and testing.
+DataJoint 2.2 introduces **isolated instances**, **thread-safe mode**, and **graph-driven diagram operations** for applications that need multiple independent database connections, explicit cascade control, and operational use of the dependency graph.
```
dimitri-yatsenko: Restructure this sentence. Thread-safe mode is for applications that handle multiple connections but graph driven operations are for general use, even for single-threaded operations.

src/explanation/whats-new-22.md
```diff
@@ -201,9 +201,72 @@ class MyTable(dj.Manual):
...
+restricted.delete(prompt=False)
+```
+
+This is valuable when working with unfamiliar pipelines, large datasets, or multi-schema dependencies where the cascade impact is not immediately obvious.
+
+### Two Propagation Modes
+
+The diagram supports two restriction propagation modes designed for fundamentally different tasks.
+
+**`cascade()` prepares a delete.** It takes a single restricted table expression, propagates the restriction downstream through all descendants, and **trims the diagram** to the resulting subgraph — ancestors and unrelated tables are removed entirely. Convergence uses OR: a descendant row is marked for deletion if *any* ancestor path reaches it, because if any reason exists to remove a row, it should be removed. `cascade()` is one-shot and is always followed by `preview()` or `delete()`.
```
dimitri-yatsenko, line 241: Here include a description of how cascade behaves when it encounters a part table whose master is not yet included in the cascade.

src/explanation/whats-new-22.md
```diff
@@ -201,9 +201,72 @@ class MyTable(dj.Manual):
...
+
+**`restrict()` selects a data subset.** It propagates a restriction downstream but **preserves the full diagram**, allowing `restrict()` to be called again from a different seed table. This makes it possible to build up multi-condition subsets incrementally — for example, restricting by species from one table and by date from another. Convergence uses AND: a descendant row is included only if *all* restricted ancestors match, because an export should contain only rows satisfying every condition. After chaining restrictions, use `prune()` to remove empty tables and `preview()` to inspect the result.
```
dimitri-yatsenko, line 245: Also, describe how part tables are restricted when reached through a path that did not include its master.

src/explanation/whats-new-22.md
```diff
@@ -201,9 +201,72 @@ class MyTable(dj.Manual):
...
+
+The two modes are mutually exclusive on the same diagram. This prevents accidental mixing of incompatible semantics — a delete diagram should never be reused for subsetting, and vice versa.
```
dimitri-yatsenko: Helpfully, datajoint will raise an error when user attempts to mix cascade and restrict or when he attempts to cascade more than once.

src/how-to/delete-data.md
```diff
@@ -189,8 +189,43 @@ count = (Subject & restriction).delete(prompt=False)
 print(f"Deleted {count} subjects")
 ```
 
+## Diagram-Level Delete
+
+!!! version-added "New in 2.2"
+    Diagram-level delete was added in DataJoint 2.2.
+
+For complex scenarios — previewing the blast radius, working across schemas, or understanding the dependency graph before deleting — use `dj.Diagram` to build and inspect the cascade before executing.
```
dimitri-yatsenko: can the same thing be accomplished with table.delete(dry_run=True)?

src/reference/specs/diagram.md
```diff
@@ -117,6 +117,182 @@ dj.Diagram(Subject) + dj.Diagram(analysis).collapse()
...
+
+**`part_integrity` values:**
+
+| Value | Behavior |
+|-------|----------|
+| `"enforce"` | Error if parts would be deleted before masters |
+| `"ignore"` | Allow deleting parts without masters |
+| `"cascade"` | Propagate restriction upward from part to master, then re-propagate downstream to all sibling parts |
+
+With `"cascade"`, the restriction flows **upward** from a part table to its master: the restricted part rows identify which master rows are affected, those masters receive a restriction, and that restriction propagates back downstream through the normal cascade — deleting the entire compositional unit (master + all parts), not just the originally matched part rows.
```
dimitri-yatsenko, line 156: If I understand this correctly, this may result in a restriction on the part table with a subquery in its where clause that references the same part table. This may produce an error in MySQL when attempting to delete from this table. Do we have a test that covers this part-to-master propagation in delete?

src/reference/specs/diagram.md
```diff
@@ -117,6 +117,182 @@ dj.Diagram(Subject) + dj.Diagram(analysis).collapse()
...
+restricted = diag.cascade(Session & {'subject_id': 'M001'})
+```
+
+### `restrict()`
+
+```python
+diag.restrict(table_expr)
+```
+
+Select a subset of data for export or inspection. Starting from a restricted table expression, propagate the restriction downstream through all descendants using **AND** semantics — a descendant row is included only if *all* restricted ancestors match. The full diagram is preserved (ancestors, unrelated tables) so that `restrict()` can be called again from a different seed table, building up a multi-condition subset incrementally.
```
dimitri-yatsenko, line 170: How does the restriction of parts restrict their masters when not reached through the master? Should part integrity be accounted in diagram restriction as well?

src/reference/specs/diagram.md
```diff
@@ -117,6 +117,182 @@ dj.Diagram(Subject) + dj.Diagram(analysis).collapse()
...
+restricted = (diag
+    .restrict(Subject & {'species': 'mouse'})
+    .restrict(Session & 'session_date > "2024-01-01"'))
+```
+
+### `delete()`
+
+```python
+diag.delete(transaction=True, prompt=None, dry_run=False)
+```
```
dimitri-yatsenko: Should delete be removed or hidden from diagram and only executed from table? Generally, delete should not be called unless cascade has been applied?  Or does this also allow for unrestricted delete from all tables in the diagram? I guess this could go either way. What is the advantage of keeping the delete in dj.Diagram if it's always called with a single seed table?

src/reference/specs/diagram.md
```diff
@@ -117,6 +117,182 @@ dj.Diagram(Subject) + dj.Diagram(analysis).collapse()
...
+# {'`lab`.`session`': 3, '`lab`.`trial`': 45, '`lab`.`processed_data`': 45}
+```
+
+### `prune()`
+
+```python
+diag.prune()
+```
+
+Remove tables with zero matching rows from the diagram. Without prior restrictions, removes physically empty tables. With restrictions (`cascade()` or `restrict()`), removes tables where the restricted query yields zero rows.
```
dimitri-yatsenko: We may want to highlight that removing from the diagram only does that -- remove from the diagram view, not affecting the tables themselves.

src/explanation/whats-new-22.md
dimitri-yatsenko: Fixed. Restructured to: "introduces **isolated instances** and **thread-safe mode** for applications that need multiple independent database connections, and **graph-driven diagram operations** that replace the legacy error-driven cascade with a reliable, inspectable approach for all users."

src/explanation/whats-new-22.md
dimitri-yatsenko: Added a paragraph after cascade() describing part table behavior: with `"enforce"` (default), `delete()` raises an error if part rows would be deleted without their master; with `"cascade"`, the restriction propagates upward from part to master, then back downstream to all sibling parts — deleting the entire compositional unit.

src/explanation/whats-new-22.md
dimitri-yatsenko: Covered in the same paragraph — describes the upward propagation from part to master when the part is reached through a path that didn't include its master.

src/explanation/whats-new-22.md
dimitri-yatsenko: Added: "DataJoint raises an error if you attempt to mix `cascade()` and `restrict()`, or if you call `cascade()` more than once."

src/how-to/delete-data.md
dimitri-yatsenko: Yes — `table.delete(dry_run=True)` returns the same affected row counts without deleting anything. Added a note at the top of the section mentioning this simpler alternative before introducing the diagram-level API.

src/reference/specs/diagram.md
dimitri-yatsenko: Yes, there are tests covering this. In `test_cascading_delete.py`:

- `test_delete_parts()` — deletes a Part table (`Website`) with `part_integrity="cascade"`, triggering upward propagation to its master (`Profile`)
- `test_delete_parts_complex()` — cascades from `A` with `part_integrity="cascade"` through complex master/part relationships, verifying correct row counts
- `test_delete_parts_error()` — verifies that `part_integrity="enforce"` raises `DataJointError` when parts would be deleted without their master

On the self-referencing subquery concern: the implementation uses `(master_ft.proj() & child_ft.proj()).to_arrays()` to materialize the restriction into a concrete tuple list before applying it to the master. This avoids a live subquery referencing the same table. The `propagated_edges` and `visited_masters` sets prevent the restriction from cycling back to the part table on subsequent passes. That said, this is worth monitoring — if the tuple list is very large, it could hit MySQL packet limits. We could add a targeted test for that edge case.

src/reference/specs/diagram.md
dimitri-yatsenko: In `cascade()` mode: when a part table is reached through a path that did not include its master, the `part_integrity` setting controls the behavior:

- `"enforce"` (default): the cascade proceeds, but `delete()` runs a post-check — if part rows were deleted without the master also being deleted, it rolls back and raises an error.
- `"cascade"`: the restriction propagates upward from the part to its master (using `to_arrays()` to materialize the join), then back downstream to all sibling parts.
- `"ignore"`: the part is deleted without checking the master.

For `restrict()` mode: `part_integrity` is not currently applied. `restrict()` is designed for subsetting/export, not delete, so the master-part integrity constraint is less relevant — you are selecting data, not removing it. If we wanted to support "include the master whenever any part matches" in restrict mode, that would be a feature addition. Worth discussing but probably not needed for the initial release.

src/reference/specs/diagram.md
dimitri-yatsenko: `Diagram.delete()` requires `cascade()` first — it raises `DataJointError("No cascade restrictions applied. Call cascade() first.")` if called on an unrestricted diagram. So it cannot do an unrestricted delete of all tables.

The advantage of keeping `delete()` on `Diagram` rather than only on `Table`:

1. **Multi-schema cascades**: `Table.delete()` constructs a diagram from `_from_table(self)`, which only includes the table and its descendants. If you need to cascade across schemas or from a specific subgraph, the diagram API lets you construct the graph first with `dj.Diagram(schema1) + dj.Diagram(schema2)`, then `cascade()`.
2. **Preview workflow**: The diagram API enables `cascade() → preview() → delete()` as a three-step pattern. `Table.delete(dry_run=True)` provides the same counts but the diagram also supports visualization of the cascade subgraph.
3. **Future operations**: `restrict() → prune() → export()` — the diagram is the foundation for data subsetting, which does not start from a single table.

That said, for the common case of "delete rows from one table", `Table.delete()` is simpler and delegates to `Diagram` internally. The diagram-level API is the power-user interface.

src/reference/specs/diagram.md
dimitri-yatsenko: Good call. Updated the `prune()` description to: "Remove tables with zero matching rows from the diagram view. This only affects the diagram object — no tables or data are modified in the database."

src/reference/specs/diagram.md
dimitri-yatsenko: Done. `delete()` and `drop()` have been removed from Diagram's public API. All mutation logic (transaction management, SQL execution, prompts) now lives in `Table.delete()` and `Table.drop()`. Diagram remains purely a graph computation and inspection tool: `cascade()`, `restrict()`, `preview()`, `prune()`.

The design docs (`docs/design/restricted-diagram.md` and `docs/design/thread-safe-mode.md`) have also been removed from datajoint-python — their content is captured in the datajoint-docs specs.

ttngu207 (approved):

esutlie (approved):

PR #162, opened 2026-04-29
docs: hidden attributes are platform-only; clarify users cannot declare them

dimitri-yatsenko:
## Summary

Reworks `reference/specs/table-declaration.md` §3.4 to reflect the design decision in datajoint/datajoint-python#1441: user-defined hidden attributes are not supported, and the parser now returns a clear \`DataJointError\` (instead of a cryptic \`pyparsing.ParseException\`) when a definition uses a leading-underscore name.

## What changed in §3.4

- **Lead paragraph** states up-front that user-defined hidden attributes are not supported, and shows the exact error message users will see.
- **Platform-managed table** (`_job_start_time`, `_job_duration`, `_job_version`, `_singleton`) is preserved — these are the actual hidden columns users encounter in fetch results, joins, and describe output.
- **Behavior matrix** is preserved — it's still the canonical reference for how platform-managed hidden columns interact with every public API surface.
- **"Why users can't declare them"** paragraph added: no public-API write path, no `describe()` round-trip, silent filtering on dict restrictions and `insert(ignore_extra_fields=True)`.
- The previous "User-defined hidden attributes" subsection and the `_params_hash`-as-hidden example are **removed**. Replaced with a regular-attribute example showing the recommended pattern: declare `params_hash` as a regular column, use `proj()` at the call site if visibility control is needed.

## Companion code PR

datajoint/datajoint-python#1441 — replaces the cryptic \`pyparsing.ParseException\` with the helpful \`DataJointError\` and adds a unit test asserting that \`compile_attribute\` rejects leading-underscore names with the new message.

## Test plan

- [x] \`python scripts/gen_llms_full.py\` regenerates cleanly.
- [ ] Build the docs site locally and confirm §3.4 renders correctly.

### DIFF
--- a/src/reference/specs/table-declaration.md
+++ b/src/reference/specs/table-declaration.md
@@ -197,9 +197,12 @@ class TaskParams(dj.Manual):
 | Natural-join namesake matching | Excluded |
 | Dict restriction `Table & {"_name": value}` | Silently ignored |
 | String restriction `Table & "_name = ..."` | Included (passes to SQL) |
-| `insert()`, `insert1()`, `update1()` | Rejected — key not in heading |
-| `describe()` / reverse-engineered definition | **Excluded** — see caveat below |
+| `insert()`, `insert1()`, `update1()` | Rejected — see write caveat below |
+| `insert(..., ignore_extra_fields=True)` | Silently dropped (key not written) |
+| `describe()` / reverse-engineered definition | **Excluded** — see round-trip caveat below |
 | `unique index (..., _name)` | Allowed |
+
+**Write caveat.** Neither `insert`/`insert1` nor `update1` accepts hidden attributes through the public API. `update1` raises `DataJointError: Attribute '_name' not found.` `insert` raises `Field '_name' not in table heading` — unless `ignore_extra_fields=True` is passed, in which case the hidden key is *silently dropped* and never written. There is currently no public-API path to populate a user-defined hidden column. Platform-managed hidden columns (the `_job_*` group) are populated by DataJoint internals via raw SQL during the `populate()` lifecycle (see `autopopulate.py`), not via the user-facing `insert`/`update1` methods. If you declare a user-defined hidden column today and need to populate it, you must do so via `connection.query()` with a raw `INSERT` or `UPDATE`, or compute it from a non-hidden column inside an `auto_populate` step.
 
 **Round-trip caveat.** `describe()` walks `heading.attributes`, so it omits hidden attributes from the regenerated definition. For platform-managed hidden columns this is harmless: re-declaring a `Computed` or `Imported` table re-injects `_job_*` automatically. For *user-defined* hidden columns (such as `_params_hash` above), the regenerated definition is incomplete — re-applying it would create a table without the hidden column. Treat `describe()` output as a starting point for review, not as a faithful round-trip when user-defined hidden columns are present.
 
--- a/src/reference/specs/table-declaration.md
+++ b/src/reference/specs/table-declaration.md
@@ -158,9 +158,16 @@ attribute_name [= default_value] : type [# comment]
 
 ### 3.4 Hidden Attributes
 
-Attributes with names starting with an underscore (`_`) are **hidden**. The hidden-attribute mechanism was designed for platform operations — bookkeeping columns DataJoint itself adds to support the data pipeline — and is filtered out of normal user-facing query results. Some hidden-attribute functionality is exposed to users as well, but the feature is not intended as a general column-hiding tool.
+Attributes with names starting with an underscore (`_`) are **hidden**. The hidden-attribute mechanism is reserved for **platform-managed** columns — bookkeeping that DataJoint itself adds to support the data pipeline — and is intentionally not exposed for user-defined attributes. Attempting to declare an attribute name with a leading underscore raises:
 
-**Platform-managed hidden attributes** are added automatically when DataJoint declares certain table types. Users do not write these in the definition:
+```text
+DataJointError: Attribute name in line "_hidden: bool" starts with an underscore.
+Names with leading underscore are reserved for platform-managed columns
+(e.g. _job_start_time, _singleton). Use a regular attribute name; if you
+need to control visibility at the call site, use proj().
+```
+
+**Platform-managed hidden attributes** are added automatically when DataJoint declares certain table types. Users do not write these in the definition; the framework injects them programmatically after parsing.
 
 | Hidden attribute | Added to | Purpose |
 |------------------|----------|---------|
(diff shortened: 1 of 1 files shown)


src/reference/specs/table-declaration.md
```diff
@@ -158,54 +158,88 @@ attribute_name [= default_value] : type [# comment]
...
+**Behavior.** The filter is implemented in `Heading.attributes`, which all visible code paths consume; raw SQL strings bypass it.
+
+| Context | Hidden attributes |
+|---------|-------------------|
+| `heading.attributes`, `heading.names`, `heading.primary_key` | Excluded |
+| `heading._attributes` (internal) | Included |
+| Table display / `repr` / `_repr_html_` | Excluded |
+| `fetch()`, `fetch1()`, `to_dicts()`, `to_pandas()` (default) | Excluded |
+| `fetch("_name")` / `fetch1("_name")` (explicit) | Included |
+| `proj("_name")` (explicit) | Included |
```
MilagrosMarin: These two rows say **Included**, but I believe the actual behavior is **Rejected** — this needs a fix.

Both `fetch("_name")` and `proj("_name")` route through `expression.py`'s `proj()`, which validates the requested name against `heading.names` — and `heading.names` is the visible-only list (`heading.py:236-237`). Passing a leading-underscore name therefore raises:

```
DataJointError: Attribute `_job_start_time` not found.
```

See `expression.py:574` for the check, and `expression.py:847` (`fetch1` → `self.proj(*proj_attrs).to_dicts(...)`) for the delegation. The integration test makes the constraint explicit at `tests/integration/test_hidden_job_metadata.py:170-172`:

```python
# Fetch hidden attributes using raw SQL since fetch() filters them
conn.query(f"SELECT _job_start_time, ... FROM {table.full_table_name}").fetchall()
```

Suggested rewording:

```
| `fetch("_name")` / `fetch1("_name")` (explicit) | Rejected (`Attribute not found`) — use raw SQL via `conn.query(...)` |
| `proj("_name")` (explicit)                      | Rejected (same reason)                                               |
```

This claim is preserved from the pre-PR §3.4 rather than introduced here, but since this PR is reframing §3.4 as the authoritative platform-behavior reference, fixing it here is the right place.

src/reference/specs/table-declaration.md
```diff
@@ -158,54 +158,88 @@ attribute_name [= default_value] : type [# comment]
...
+# Default fetch — hidden columns excluded
 results = MyTable.to_dicts()
 
-# Explicitly include hidden attributes
+# Explicit projection promotes a hidden column to visible
 results = MyTable.proj('result', '_job_start_time').to_dicts()
 
-# Or with fetch1 for single row
+# Explicit fetch by name returns hidden columns
 row = (MyTable & key).fetch1('result', '_job_start_time')
```
MilagrosMarin: Tied to the matrix comment above — these two examples won't run as written:

```python
results = MyTable.proj('result', '_job_start_time').to_dicts()  # DataJointError: Attribute `_job_start_time` not found.
row = (MyTable & key).fetch1('result', '_job_start_time')        # same: fetch1 routes through proj()
```

Suggested replacement (matches what the integration test actually does):

```python
# Default fetch — hidden columns excluded
results = MyTable.to_dicts()

# To inspect platform-managed hidden columns, query raw SQL.
# The public API (fetch / proj) intentionally filters them out.
conn = MyTable.connection
rows = conn.query(
    f"SELECT _job_start_time, _job_duration, _job_version "
    f"FROM {MyTable.full_table_name}"
).fetchall()

# String restriction works (passes through to SQL)
MyTable & "_job_start_time > '2024-01-01'"

# Dict restriction is silently dropped — does NOT filter
MyTable & {'_job_start_time': some_date}   # ⚠ ignored
```

src/reference/specs/table-declaration.md
```diff
@@ -158,54 +158,88 @@ attribute_name [= default_value] : type [# comment]
...
+| `heading.attributes`, `heading.names`, `heading.primary_key` | Excluded |
+| `heading._attributes` (internal) | Included |
+| Table display / `repr` / `_repr_html_` | Excluded |
+| `fetch()`, `fetch1()`, `to_dicts()`, `to_pandas()` (default) | Excluded |
+| `fetch("_name")` / `fetch1("_name")` (explicit) | Included |
+| `proj("_name")` (explicit) | Included |
+| Natural-join namesake matching | Excluded |
+| Dict restriction `Table & {"_name": value}` | Silently ignored |
+| String restriction `Table & "_name = ..."` | Included (passes to SQL) |
+| `insert()`, `insert1()`, `update1()` | Rejected (`Field not in table heading`) |
```
MilagrosMarin: Minor wording nit: the parenthetical `(Field not in table heading)` is approximate. Actual messages:

- `insert` / `insert1`: `KeyError("`_name` is not in the table heading")` (`table.py:1424`)
- `update1`: `DataJointError("Attribute `_name` not found.")` (`table.py:514`)

Either match one verbatim or generalize, e.g. `Rejected — name not in visible heading`. Not blocking, just so the matrix is precise.

MilagrosMarin (commented):
Thanks @dimitri-yatsenko for this thorough rework! The reframing of §3.4 as platform-only — with the upfront error message, the platform-managed table, the "why users can't declare them" rationale, and the regular-attribute alternative — is a real improvement over the previous version. Verified the companion PR datajoint/datajoint-python#1441 (merged), and the error message block in §3.4 matches the code in `declare.py:858` verbatim. ✅

Most of the behavior matrix also checks out against `datajoint-python` master:

- `heading.attributes` / `heading.names` / `heading.primary_key` exclude hidden ✅ (`heading.py:230-247`)
- `heading._attributes` includes hidden ✅ (`heading.py:204`)
- `to_dicts` / `to_pandas` default exclude ✅ (`expression.py:899`)
- Natural-join namesake matching excludes hidden ✅ (`expression.py:397-398`)
- Dict restriction silently ignored ✅ (`condition.py:392`)
- String restriction passes through to SQL ✅
- `describe()` excludes ✅ (`table.py:1233`)
- `ignore_extra_fields=True` silently drops hidden ✅ (`table.py:1443`)
- Platform-managed columns populated via raw SQL during `populate()` ✅ (`autopopulate.py:766-789`)

A few comments below — one critical accuracy issue on the `fetch("_name")` / `proj("_name")` rows of the matrix (and the corresponding example code), plus a minor wording nit on the `insert/update1` row.

dimitri-yatsenko:
Thank you @MilagrosMarin for the careful read — and especially for catching the `fetch("_name")` / `proj("_name")` claim. That was a real factual error inherited from the pre-PR §3.4 that would have misled users into writing code that raises. Equally helpful was confirming the constraint via the `tests/integration/test_hidden_job_metadata.py:170-172` raw-SQL workaround — that's exactly the citation the docs needed to point at.

All three comments addressed in `20224ae`:

1. **Matrix rows for `fetch("_name")` / `proj("_name")`** — flipped from `Included` to `Rejected (Attribute not found) — use raw SQL via conn.query(...)`.
2. **`Inspecting platform-managed hidden columns` example block** — replaced the would-have-raised `proj()`/`fetch1()` snippets with the raw-SQL pattern that mirrors the integration test.
3. **`insert`/`update1` row** — split into two rows with the verbatim error messages: `KeyError("\`_name\` is not in the table heading")` (`table.py:1424`) and `DataJointError("Attribute \`_name\` not found.")` (`table.py:514`).

Ready for another pass when you have a moment.

MilagrosMarin (approved):
Thanks @dimitri-yatsenko! Re-read §3.4 at `20224ae` and verified all three:

✅ **Matrix rows for `fetch("_name")` / `proj("_name")`** — now correctly state `Rejected (Attribute not found) — use raw SQL via conn.query(...)`. Matches the actual `proj()` validation against `heading.names` (`expression.py:574`) and the integration-test workaround.

✅ **Example block** — replaced with the `conn.query(...)` raw-SQL pattern that mirrors `tests/integration/test_hidden_job_metadata.py:170-172`. Reads cleanly and is now actually runnable.

✅ **Insert/update1 rows** — split into two rows with verbatim error messages: ``KeyError("`_name` is not in the table heading")`` (`table.py:1424`) for `insert`/`insert1`, and ``DataJointError("Attribute `_name` not found.")`` (`table.py:514`) for `update1`. Matrix is now precise.

The whole §3.4 section now reads as a clean, accurate platform-behavior reference. LGTM — approving.

MilagrosMarin (approved):

PR #174, opened 2026-05-20
ci+scripts: address review nits from #170

dimitri-yatsenko:
## Summary
Follow-ups to @MilagrosMarin's review on #170. Each of the four nits she flagged is now addressed in one self-contained PR — no notebook re-execution involved.

| # | Nit (from #170 review) | Fix here |
|---|---|---|
| 1 | `check_notebook_versions.py` not in CI | New `.github/workflows/check-notebooks.yml` runs the check on PRs that touch notebooks, `mkdocs.yaml`, or the script |
| 2 | Pre-cache failure swallowed silently | Warnings now go to `sys.stderr` (visible in CI logs) without escalating to a non-zero exit — a transient gitlab.com hiccup shouldn't block the whole run |
| 3 | `skimage` imported at module-execute time | `import skimage.data` wrapped in try/except; script degrades gracefully when scikit-image isn't installed |
| 4 | `had_banner` short-circuits after first stale match | Drop the inner `break`; collect stale versions per notebook in a set so multiple stale banners are all reported, deduped |

### Design note on #1
Milagros suggested wiring into `.github/workflows/development.yml`, but that workflow only triggers on `push: main` (it's the gh-pages deploy). Adding the check there would fire post-merge and never block a PR. A separate `pull_request`-triggered workflow that just runs the script (no Docker, no datajoint-python checkout) gives the intended fail-the-PR behavior in seconds.

## Test plan
- [ ] CI: this PR itself should run the new `Check notebooks` workflow once green (since `.github/workflows/check-notebooks.yml` is in its own paths filter)
- [ ] Local: `python scripts/check_notebook_versions.py` → `OK: 22 notebook(s) with banners are on DataJoint 2.2.x` (verified)
- [ ] Local: `python scripts/execute_notebooks.py --help` works without scikit-image installed (verified — argparse exits before import anyway, and import is now guarded)
- [ ] Manual smoke: temporarily bump a notebook's banner to `2.0.x`, run the check, confirm it's reported (not just the first match) — exits 1

### DIFF
--- a/.github/workflows/check-notebooks.yml
+++ b/.github/workflows/check-notebooks.yml
@@ -1,0 +1,19 @@
+name: Check notebooks
+on:
+  pull_request:
+    paths:
+      - 'src/tutorials/**/*.ipynb'
+      - 'src/how-to/**/*.ipynb'
+      - 'mkdocs.yaml'
+      - 'scripts/check_notebook_versions.py'
+      - '.github/workflows/check-notebooks.yml'
+jobs:
+  check-versions:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+      - name: Verify notebook DataJoint connection banners are current
+        run: python scripts/check_notebook_versions.py
--- a/scripts/check_notebook_versions.py
+++ b/scripts/check_notebook_versions.py
@@ -80,22 +80,22 @@ def main() -> int:
             notebooks.extend(p for p in d.rglob("*.ipynb") if ".ipynb_checkpoints" not in str(p))
     notebooks.sort()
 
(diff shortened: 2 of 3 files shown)


MilagrosMarin (approved):
Thanks @dimitri-yatsenko! All four follow-ups verified end-to-end:

✅ CI workflow triggers on `pull_request` with a tight path filter — your design-note pushback is right, `development.yml` only fires post-merge on gh-pages deploy.
✅ Pre-cache warnings now visible via `sys.stderr`; not escalating to non-zero exit is a reasonable trade-off.
✅ `python scripts/execute_notebooks.py --help` works locally without scikit-image installed.
✅ Multiple stale banners are now all reported, sorted, deduped — verified with a 2-banner injection test.

Also confirmed the positive case after #171: `OK: 22 notebook(s) with banners are on DataJoint 2.2.x`, exit 0.

Approving — clean follow-up.

PR #188, opened 2026-06-23
docs(#1458): SparkAdapter Codec Protocol spec + explainer

dimitri-yatsenko:
## Summary

Spec-first pair for the Renderable Codec Protocol landing in DataJoint 2.3 ([#1458](https://github.com/datajoint/datajoint-python/issues/1458)). Brought into 2.3 scope per user direction 2026-06-23 (was previously deferred to 2.4).

| File | Role |
|---|---|
| \`src/reference/specs/renderable.md\` (new) | Normative spec: Protocol signature, return-value shape constraints (Spark-native: primitives / lists / dicts), why it's a Protocol vs. abstract method, eligibility detection via \`isinstance\`, out-of-scope items, worked codec examples. |
| \`src/explanation/renderable-codecs.md\` (new) | Explainer: Bronze/Silver layer model, why \`<blob@>\` is bronze-only, design rationale (smaller OSS surface, cleaner opt-in, no plugin churn, structural typing), decision guide for choosing codecs in a new pipeline. |
| \`mkdocs.yaml\` | Nav entries under Reference → Specifications → Type System, and Concepts → Storage. |

## Why spec-first

The Protocol itself is tiny (~10 lines: a \`@runtime_checkable Protocol\` declaration). The design conversation in the issue body settled the shape after [#1457](https://github.com/datajoint/datajoint-python/issues/1457) (the earlier abstract-method-on-Codec framing) was rejected. Locking the spec now gives:

- The downstream consumer ([\`datajoint-databricks\`](https://github.com/datajoint-company/datajoint-databricks)) a stable contract to build the silver-layer publish pipeline against.
- Codec plugin authors (current and future) a clear opt-in target.
- The implementation PR a small, well-scoped diff to land against.

## Marked draft

Stays draft until the matching implementation PR opens in \`datajoint-python\` — same pattern as the provenance trinity spec ([#183](https://github.com/datajoint/datajoint-docs/pull/183)) before #1471 landed against it.

## Sequencing

Independent of the provenance trinity (no code overlap). Can land in parallel with T2.2 implementation work.

1. This PR (spec) — review while implementation is drafted.
2. datajoint-python implementation PR — adds \`src/datajoint/rendering.py\` with the Protocol, re-exports as \`dj.Renderable\`, adds unit tests.
3. Flip this PR from draft to ready alongside the implementation PR.

## Test plan

- [ ] \`mkdocs serve\` renders both pages under the new nav groups
- [ ] Cross-links resolve (codec-api.md, custom-codecs.md explainer, the issue, #1457)
- [ ] Examples use core DataJoint types per project convention
- [ ] Reviewers can sketch a plugin codec from the worked examples without ambiguity

### DIFF
--- a/mkdocs.yaml
+++ b/mkdocs.yaml
@@ -24,6 +24,7 @@ nav:
       - Storage:
           - Type System: explanation/type-system.md
           - Custom Codecs: explanation/custom-codecs.md
+          - Renderable Codecs: explanation/renderable-codecs.md
       - Operations:
           - PostgreSQL CDC and Replica Identity: explanation/postgresql-cdc-replication.md
   - Tutorials:
@@ -115,6 +116,7 @@ nav:
               - Types: reference/specs/type-system.md
               - Codec API: reference/specs/codec-api.md
               - NPY Codec: reference/specs/npy-codec.md
+              - Renderable Codec Protocol: reference/specs/renderable.md
               - Storage Adapter API: reference/specs/storage-adapter-api.md
           - Data Operations:
               - Data Manipulation: reference/specs/data-manipulation.md
--- a/mkdocs.yaml
+++ b/mkdocs.yaml
@@ -24,7 +24,7 @@ nav:
       - Storage:
           - Type System: explanation/type-system.md
           - Custom Codecs: explanation/custom-codecs.md
-          - Renderable Codecs: explanation/renderable-codecs.md
+          - Spark Adapters: explanation/spark-adapters.md
       - Operations:
           - PostgreSQL CDC and Replica Identity: explanation/postgresql-cdc-replication.md
   - Tutorials:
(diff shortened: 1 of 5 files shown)


ttngu207 (approved):

MilagrosMarin (approved):
Verified against the shipped code in #1472. The Protocol signature, module path, `dj.SparkAdapter` re-export, and blob/hash non-eligibility all match the implementation exactly, and the line-87 caveat about `@runtime_checkable` checking method existence but not signature is exactly the observation I flagged independently on #1472 — good that this spec surfaces it. The design rationale for Protocol vs. abstract method reads well, and the out-of-scope enumeration is precise.

One thing worth a sanity check before merge: the `FloatArrayCodec.get_dtype` example on line 112 returns `"<hash@>"` when `is_store=True`, but the built-in `BlobCodec.get_dtype` on master returns `"<hash>"` (no `@` — the `@` modifier gets composed by the framework at declaration time, not returned by the codec). If the intent is to mirror built-in conventions, plugin authors copy-pasting the example may see divergent behavior. Someone deeper in the codec-chain semantics should confirm.

Two cosmetic residuals from the Renderable → SparkAdapter rename: the PR summary table still lists `renderable.md` / `renderable-codecs.md` (files are `spark-adapter.md` / `spark-adapters.md`), and the "Stays draft until the matching implementation PR opens" framing is out of date since #1472 is now open and approved. Nav placement, cross-links, and the explainer's Bronze/Silver framing are all clean.

Approving.

PR #193, opened 2026-07-01
docs(deploy): note set_replica_identity partial-failure semantics

dimitri-yatsenko:
One-line addition to the `set_replica_identity` spec (`deploy-operations.md`), following the #1466 implementation landing on `datajoint-python` master.

The spec's **Behavior** section documented idempotency but omitted the failure mode: the operation is **not atomic** — tables are altered one at a time, so if execution raises on table _N_ of _M_, the first _N_−1 ALTERs have already committed and the exception propagates without returning a summary. This matches the behavior already documented in `deploy.py`'s docstring; re-running converges since each ALTER is idempotent.

Surfaced during a spec-vs-implementation consistency check of datajoint/datajoint-python#1466.


MilagrosMarin (approved):
Matches `deploy.py`'s "Partial-failure semantics" docstring on master exactly, and closes the spec-vs-implementation gap from my #1466 review point #7. Approving.

PR #195, opened 2026-07-04
docs: post-audit reconciliation — cascade U3/limitations + trace cross-refs

dimitri-yatsenko:
Post-2.3 audit reconciliation for the **cascade** and **trace** specs.

## Contents
- **`cascade.md`** — corrects the U3 upward rule (`child.proj(*attr_map.keys())`), notes the enforce post-check is table-level, and adds a **Limitations** section (single FK path, materialization cost, empty-match sentinel, enforce granularity).
- **`diagram.md`** — documents `Diagram.trace()` (the upstream mirror of `cascade()`) and the `part_integrity` preview semantics.
- **`autopopulate.md`** — adds the §4.3.1 `self.upstream` (pre-restricted ancestor access) section.
- **`whats-new-23.md`** — minor wording.

## Scope note
Originally this PR also documented the `strict_provenance` runtime guardrail (a how-to section in `run-computations.md` and enforcement-limit edits in `provenance.md`). That guardrail is being **retired** in datajoint-python #1489, and its spec page is removed in #199 (which deletes `provenance.md` and adds `trace.md`). Those parts have been dropped here to stay coherent.

⚠️ **Depends on #199 for link reconciliation.** The retained `cascade.md` / `diagram.md` / `autopopulate.md` still cross-link to `provenance.md` (`#1-diagramtracetable_expr`, `#2-selfupstream-inside-make`). Those anchors resolve against `main` today, but #199 deletes `provenance.md` in favor of `trace.md` — so after #199 merges these links must be repointed to `trace.md`. Recommend landing #199 first (or repointing as part of it).


src/reference/specs/autopopulate.md
```diff
@@ -317,7 +317,28 @@ def make(self, key):
 
 **Upstream-only convention:** Inside `make()`, fetch only from tables that are strictly upstream in the pipeline—tables referenced by foreign keys in the definition, their ancestors, and their part tables. This ensures reproducibility: computed results depend only on their declared dependencies.
 
-This convention is not currently enforced programmatically but is critical for pipeline integrity. Some pipelines violate this rule for operational reasons, which makes them non-reproducible. A future release may programmatically enforce upstream-only fetches inside `make()`.
+As of DataJoint 2.3, this convention is checked (opt-in) at runtime: setting `dj.config["strict_provenance"] = True` makes reads from undeclared tables and writes to tables other than `self` (and its Parts) raise `DataJointError` inside `make()`. The check is a best-effort development guardrail; see the [Provenance Specification](provenance.md) for the enforcement model and its documented limits.
```
dimitri-yatsenko: Reading from any table that is not part of the "trace" (all tables above in the DAG and their part tables) will raise an error, not "undeclared tables."

MilagrosMarin (dismissed):
Thanks @dimitri-yatsenko — good post-audit reconciliation. Noting `324e64e` already dropped the strict_provenance prose (matching your review-push note on the batch), so this PR is now the cascade/trace reconciliation part only. Verified.

The **U3 correction** matches shipped code exactly (`src/datajoint/diagram.py`: `parent_item = child_ft.proj(*attr_map.keys())`), and Worked Example 1's walk-through now reflects it. The **F1/U1 "non-empty" precondition** closes a gap I would have flagged: the old rule table left the empty-attrs case ambiguous; the "empty attribute set takes rule 3" note resolves it.

New **Limitations** section is honest and matches shipped behavior: single FK path (`nx.shortest_path` picking one arm), materialization cost bounded by distinct master rows (same reason preview pays the cost too), empty-match sentinel (matches the `[False]` branch in `_propagate_part_to_master`), and enforce granularity (table-level, not row-level — with the false-negative/false-positive framing spelled out). The `part_integrity="enforce"` table row now cross-references it.

`diagram.md` — good correction: *"Error if parts would be deleted before masters"* was misleading (the preview itself never errors) and now correctly attributes enforcement to `Table.delete()`'s post-check. `Diagram.trace()` documented alongside `cascade()` with the same shape. `autopopulate.md` §4.3.1 covers the four `self.upstream` semantic properties (lifecycle, laziness, tripartite availability, ancestors-only scope) matching what I verified when reviewing #1473.

**Sequencing note — worth confirming beyond the PR body's own flag.** The retained `cascade.md` / `diagram.md` / `autopopulate.md` references to `provenance.md` will resolve at runtime via #199's redirect (the anchor slugs `#1-diagramtracetable_expr` and `#2-selfupstream-inside-make` survive the rename since `trace.md` retains those section headers). But `mkdocs build --strict` resolves links against source-tree file paths, not redirects — so the `provenance.md` references would fail linkcheck the moment #199 merges. Cleanest path is to repoint the three refs to `trace.md` as part of merging #199, or land them simultaneously. Either way, the flag in the PR body is on the money.

Approving.

MilagrosMarin (dismissed):

MilagrosMarin (approved):

PR #196, opened 2026-07-06
docs(storage): settings-driven section prefixes, per-schema hash paths, GC semantics

dimitri-yatsenko:
Documentation counterpart to **datajoint/datajoint-python#1479** (in review), which aligned the code to the model these docs already described: the per-store `hash_prefix` / `schema_prefix` / `filepath_prefix` settings **control the storage layout** and are consumed by writers, garbage collection, and `<filepath@>` validation from the same spec values (no component-level fallbacks).

## The load-bearing point (per maintainer review)
**Hash-addressed paths embed the schema name** — `{hash_prefix}/{schema}/{hash}` — which:
- scopes deduplication **per schema** (the type-system spec previously claimed "across the entire project" — corrected);
- lets GC attribute every stored object to a schema;
- and, because storage-side listing covers the whole section, means `collect()` **must be given every schema that uses the store** — the GC how-to's old note ("you only need to scan schemas that share the same database") asserted the opposite and is replaced with an explicit warning. Per-schema attribution is also what makes dropped-schema debris reclaimable.

## Other corrections
- **Layout parameterization:** all path illustrations now show the configurable sections; schema-addressed pages no longer show the root-level `{schema}/…` layout (that's the ≤2.3.0 legacy layout, noted as such; GC lists both).
- **Write tokens:** schema-addressed filenames are `{field}_{token}{ext}` — several pages showed a dot-separated `{field}.{token}.{ext}` form that never matched the code.
- **GC how-to:** token-per-write versioning, folder objects + `.manifest.json` sidecars with coverage-based orphan matching, and the `filepath_prefix` never-touched guarantee.
- **object-store-configuration spec:** new normative *Consumption* paragraph (single source of values, defaults applied to every spec including plugin protocols), the hash-paths-embed-schema note, and the changing-prefixes-on-populated-store caveat (also added as a warning in configure-storage).

## Sequencing
Merge **after** datajoint/datajoint-python#1479 — these pages describe the post-fix behavior. No file overlap with the open #195.

### DIFF
--- a/src/explanation/type-system.md
+++ b/src/explanation/type-system.md
@@ -266,14 +266,14 @@ result = np.mean(ref)  # Downloads automatically
 
 ### `<object@>` — Schema-Addressed Storage
 
-Schema-addressed storage for files and folders. Path mirrors the database structure: `{schema}/{table}/{pk}/{attribute}`.
+Schema-addressed storage for files and folders. Path mirrors the database structure within the store's schema-addressed section: `{schema_prefix}/{schema}/{table}/{pk}/{attribute}_{token}` (`schema_prefix` defaults to `_schema`).
 
 ```python
 class RecordingAnalysis(dj.Computed):
     definition = """
     -> Recording
     ---
-    results : <object@>       # Stored at {schema}/{table}/{pk}/results/
+    results : <object@>       # Stored at _schema/{schema}/{table}/{pk}/results_{token}/
     """
 ```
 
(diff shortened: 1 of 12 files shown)


dimitri-yatsenko:
Added a commit removing `dj.gc.format_stats(stats)` from the examples (garbage-collection.md, manage-large-data.md). That helper is being removed in datajoint/datajoint-python#1479 — it presumed a single display format and the library never called it. Examples now inspect the plain stats dict directly, and note that `collect(dry_run=True)` reports counts only (use `scan()` for per-item paths and reclaimable bytes).

dimitri-yatsenko:
Added a commit updating the GC docs to the store-bound `GarbageCollector` class (datajoint/datajoint-python#1479): examples use `dj.gc.GarbageCollector(store=...).scan(...)`/`.collect(...)`; the store binds at construction (no `store_name=` param). Two notable content changes: the **Multiple Schemas** guidance is now *inverted* — per-schema scoping makes any subset safe, so the old "always pass every schema or lose data" warning is replaced with a per-schema-attribution note; and the removed combined totals (`orphaned`/`orphaned_bytes`/`referenced`/`stored`) are dropped from the statistics reference (readers sum the per-section fields).

ttngu207 (approved):

PR #46, opened 2023-04-03
Update for Year 3

kabilar:
- [x] Remove `Works` page since it says `Coming soon` and there are no immediate plans to add documentation.  Link directly to https://works.datajoint.com/.
- [x] Update layout for `Partnership` pages
- [x] Update Elements `Developer Guide` - Remove `jupytext` since we are no longer using
- [x] Update `Core` page - Add link to the `datajoint-tutorials` repo for GitHub Codespaces
- [x] Update `Core` page - Remove hyperlinks to improve readability
- [x] Update `Publications` page
  - [x] Add new publications, and remove duplicate items (i.e. bioRxiv) for manuscripts that have now been published
  - [x] Fix incorrect link
  - [x] Update html syntax to markdown syntax for external links
- [x] Move `Dissemination` plan to separate page, and update links
- [x] Simplify `Support` page
- [x] Update `Team` page
- [x] Update `Events` page
- [x] Add upcoming Elements
- [x] Remove Google Analytics environment variable since we are now using HubSpot

PR #198, opened 2026-07-13
Document the make() reproducibility contract (#197)

### LINKED ISSUE #197: Document the make() reproducibility contract (the five rules)
The **make() reproducibility contract** — the rule set that makes a computed row's provenance intrinsic — is a set of fundamental framework design principles that is not yet written down in the documentation. This issue proposes documenting it so users can learn to write provenance-respecting pipelines.

## The five rules to document

1. **Populate-only.** Computed/Imported tables are never inserted into directly; rows are produced only by `make()` via `populate()` (enforced by the insert guard).
2. **One entity per call, computed in isolation.** One `make(key)` computes exactly one master entity (plus its Part rows); jobs share no state and assume no ordering — no reading peer rows of the same table. This is what makes `populate()` embarrassingly parallel and each entity's provenance self-contained.
3. **Read only upstream, restricted to the job.** A `make(key)` may fetch only from declared ancestors reachable from `key` — the `self.upstream` cone. The violation is reading *outside the cone* (a non-ancestor, or entities under a different key).
4. **Write the result to `self` + its Parts; extrinsic writes are allowed.** The derived result goes to `self`/Parts atomically. A `make()` may also write to other destinations (Manual tables, files, ELNs, knowledge graphs) *provided it supplies external-source provenance to each destination* — the fan-out pattern.
5. **No other result-affecting input.** `make_kwargs` may orchestrate (batching, parallelism, logging) but must never parameterize the result; anything that changes *what* is computed must enter through a declared upstream table.

One-line frame: *read from `self.upstream`, write to `self` — the read/write boundary is the provenance boundary.*

## Suggested placement
- Reference spec: extend `src/reference/specs/provenance.md` (or `autopopulate.md`) with a "make() reproducibility contract" section listing the five rules.
- Explanation: a short conceptual note in `src/explanation/computation-model.md` on why the contract yields intrinsic provenance.
- Describe the enforcement arc honestly: convention → the 2.3 read surface (`Diagram.trace` / `self.upstream`) makes the contract easy to follow and inspect. The framework does not itself *enforce* the contract at runtime.

## Notes
- The opt-in `strict_provenance` flag is being retired ([datajoint-python #1488](https://github.com/datajoint/datajoint-python/issues/1488)); document the rules as the contract, not the flag.
- Keep "code provenance" (which code produced a data entity) distinct from industry "software provenance" (SLSA/SBOM/artifact supply-chain).

dimitri-yatsenko:
Closes #197.

Documents the **make() reproducibility contract** — the five rules that make a computed row's derivation self-contained and reproducible — in the relational workflow model's own vocabulary (computed results, reproducibility, the FK DAG, table tiers, the read/write boundary).

## Changes

- **`reference/specs/autopopulate.md`** — new **§4.3 "The make() reproducibility contract"** stating all five rules (populate-only; one entity per call in isolation; read only the key-restricted upstream cone; write to `self`/Parts with fan-out writes carrying their source; no result-affecting `make_kwargs`), plus the one-line **read from `self.upstream`, write to `self`** frame and an honest enforcement note (2.3 provides the `self.upstream` / `Diagram.trace` read surface that makes the contract easy to follow and inspect; the framework does not enforce it at runtime). Reconciles the existing "Upstream-only convention" and `make_kwargs` anti-pattern notes to cross-reference the contract, fixes the stale "a future release may programmatically enforce…" wording, and renumbers §4.4–4.6.
- **`explanation/computation-model.md`** — a "Why the contract matters" note under "The `make()` Contract": the read/write boundary as the organizing idea, tied to the Computed (reproducible from within the pipeline) vs. Imported (reads untracked sources, records source identity) tier distinction.

## Notes

- Deliberately avoids the term "provenance" — this documents the contract in the model's native language. (The separate retirement of the `strict_provenance` framing is handled in its own PR, matching datajoint-python's strict_provenance removal.)
- Cross-links and anchor slugs verified; `strict_provenance` is not referenced.

### DIFF
--- a/src/explanation/computation-model.md
+++ b/src/explanation/computation-model.md
@@ -47,6 +47,34 @@ DataJoint guarantees:
 - Failed computations can be retried
 - Parallel execution is safe
 
+### Why the contract matters
+
+These guarantees hold because a well-behaved `make()` observes a small set of
+rules — the **make() reproducibility contract** (specified in full in the
+[AutoPopulate reference](../reference/specs/autopopulate.md#43-the-make-reproducibility-contract)).
+The organizing idea is a single **read/write boundary**: a `make(key)` reads only
+from its declared upstream dependencies, restricted to the current `key`, and
+writes only to `self` and its Part tables. Because each call sees a fixed,
+key-restricted slice of the pipeline and shares no state with other calls, every
+computed row is *self-contained* — produced by one `make()` over a specific set
+of declared inputs — which is exactly what makes results reproducible and
+`populate()` safe to run in parallel.
+
+This boundary is why the auto-populated tiers split into two:
+
+- **Computed** tables derive entirely from other pipeline tables. Every input is
+  itself tracked under referential integrity, so a Computed result is
+  reproducible from within the pipeline alone — re-running `make()` on the same
+  upstream data yields the same result.
+- **Imported** tables read a source the pipeline does *not* track (a file, an
+  instrument, an API). They cannot be reproduced from the pipeline alone, so an
+  Imported `make()` is responsible for recording the source's identity (path,
+  checksum, endpoint, external record ID) alongside the row.
+
+**Manual** and **Lookup** tables are not auto-populated; they are the entry points
+where a pipeline's data originates and where every downstream `make()` chain
+ultimately begins.
+
 ## Key Source
 
 The **key source** determines what needs to be computed:
--- a/src/reference/specs/autopopulate.md
+++ b/src/reference/specs/autopopulate.md
@@ -298,7 +298,28 @@ def make(self, key):
 - **Atomic**: Runs within a transaction—all or nothing
 - **Self-contained**: Should not depend on external state that changes
 
-### 4.3 Accessing Source Data
+These requirements are the surface of a broader rule set — the make() reproducibility contract (§4.3) — that keeps every computed row traceable to its declared inputs.
+
+### 4.3 The make() reproducibility contract
+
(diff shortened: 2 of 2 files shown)


MilagrosMarin (approved):
Thanks @dimitri-yatsenko — the reproducibility contract in the model's own vocabulary is the right level. Verified the runtime claims:

- **Rule 1 (populate-only, direct inserts rejected)** matches `src/datajoint/table.py:830` (`_allow_insert = False` on `AutoPopulate` gates the insert path with the exact "can only be done inside its make method" error).
- **Rule 3's ergonomic surface** — `self.upstream` (shipped in 2.3 via #1473) and `Diagram.trace` (#1471) — is cited correctly, and replacing the stale "*a future release may programmatically enforce*" with the honest "*the framework does not enforce it at runtime*" note is the right posture, especially after the `strict_provenance` retirement.
- **Renumbering §4.4/4.5/4.6** and the internal anchors (`#43-…`, `#44-…`, `#46-…`) are consistent across both files.

The one-line frame — **"read from `self.upstream`, write to `self`"** — is memorable and lands the contract in a single sentence. The Computed vs Imported split in `computation-model.md` (Computed = reproducible from within the pipeline; Imported = reads untracked source and must record source identity) is genuinely useful pedagogy that turns the read/write boundary into the organizing idea.

One precision quibble on Rule 3: *"the declared ancestors reachable by following the foreign keys embedded in the primary key"* is narrower than what `self.upstream` actually exposes. `Diagram.trace` walks `nx.ancestors`, which includes ancestors reached via both primary- and secondary-FK edges — so `self.upstream[T]` returns any declared ancestor, not just PK-ancestors. Softening to *"the declared ancestors reachable by following declared foreign keys"* (or just "the declared ancestors") avoids under-specifying the read surface.

Approving.

PR #199, opened 2026-07-13
Retire strict_provenance docs; drop provenance terminology

dimitri-yatsenko:
Documentation counterpart to datajoint/datajoint-python#1489 (retire the `strict_provenance` runtime guardrail), and alignment with the decision to keep **provenance terminology out of DataJoint's own documentation**.

DataJoint implements the **relational workflow model** — data integrity, consistency, lineage, and reproducibility. "Provenance" is associated with a different class of systems; DataJoint's provenance *compatibility* features live in the platform and are documented separately (`datajoint-provenance`). So the public docs use the model's native vocabulary: lineage, reproducibility, traceability.

> **Stacked on #198** (base branch `docs/197-make-contract`). This PR links to the *make() reproducibility contract* section added there. Merge #198 first; GitHub will retarget this PR to `main` automatically.

## Changes

**Retire `strict_provenance`:**
- Rename `reference/specs/provenance.md` → **`reference/specs/trace.md`** and rewrite it as the spec for the upstream read surface (`Diagram.trace` + `self.upstream`) — the `strict_provenance` section (§3), config docs, examples, and references are removed; the "read/write boundary" is reframed in lineage/reproducibility terms.
- `about/whats-new-23.md`: reframe the "provenance trinity" as the **`trace` + `self.upstream` pair**, drop the `strict_provenance` subsection and adoption step, retarget links.
- `mkdocs.yaml`: nav "Provenance Trace & Strict Provenance" → **"Upstream Trace"**.

**Drop "provenance" terminology** across the conceptual and tutorial pages (`index`, `about/index`, `relational-workflow-model`, `normalization`, `comparison-to-workflow-languages`, `schema-as-workflow-specification`, `data-pipelines`, `faq`, and three tutorial notebooks) — replaced with "lineage" / "reproducibility" / "traceability" per context.

## Verification

- `grep` confirms **no "provenance" terminology** remains in the framework documentation, and **no links** point to the old `specs/provenance.md` path. The sole remaining occurrence under `src/` is the platform-architecture label in `images/dj-platform.svg` ("Reproducibility • Provenance • Integrity • Collaboration") — intentionally kept: it describes the **DataJoint platform**, which *builds* provenance capabilities, whereas the framework docs describe the open-source core's relational workflow model, where "lineage" is the apt term.
- The renamed page and all retargeted cross-links resolve; anchor slugs verified. (`llms-full.txt` is a generated, untracked build artifact — regenerated on build.)

## Published URL

[...]

PR #200, opened 2026-07-13
Distinguish the roles of Manual vs Lookup tables

dimitri-yatsenko:
The docs describe all four table tiers but never state, side by side, how **Manual** and **Lookup** differ or when to choose each. This adds that distinction, in the relational workflow model's own vocabulary (no provenance framing).

## The distinction

Both are **entry-point** tiers — their rows are *entered*, not computed by a `make()`. They differ in **where the rows come from**:

- **Manual** — rows arrive at **runtime** from outside the pipeline (a person, a form, a LIMS, an instrument, an import). Project/experiment-specific; the pipeline's origin points.
- **Lookup** — rows are **part of the schema definition**, declared in code via `contents` and versioned with the table. Identical across deployments until the code changes (parameter sets, method definitions, controlled vocabularies, enumerations).

Quick test: *where does a row come from?* Committed `contents` → **Lookup**; runtime entry → **Manual**. Common misclassification: populating a "Lookup" from a dashboard form — that data is runtime-entered and belongs in **Manual**.

## Changes

- **`explanation/relational-workflow-model.md`** — new **"Manual vs. Lookup"** subsection under the tier table, and sharper Manual/Lookup role cells.
- **`how-to/define-tables.md`** — a practical **"Manual or Lookup?"** note beside the Lookup-`contents` example, plus clearer tier-table purposes.
- **`reference/specs/table-declaration.md`** — sharpen the tier "Purpose" cells.

Examples reuse the page's existing worked pipeline (`Mouse`/`Session`/`Scan`, `SegmentationParam`) for consistency.

## Notes

- Based on `main`, independent of the other open docs PRs (#198, #199). It touches `relational-workflow-model.md` in a different region than #199's "provenance" sweep, so they merge cleanly in either order.
- Distilled from internal tier-role discussion, deliberately without provenance terminology.

### DIFF
--- a/src/explanation/relational-workflow-model.md
+++ b/src/explanation/relational-workflow-model.md
@@ -134,14 +134,40 @@ Tables are classified into tiers by data-entry mode:
 
 | Tier | Role | `make()` |
 |------|------|----------|
-| **Manual** | Receive direct user entry | No |
-| **Lookup** | Hold reference data | No |
+| **Manual** | Rows entered at runtime from outside the pipeline (people, forms, instruments, imports) | No |
+| **Lookup** | Reference rows defined in the schema itself via `contents` | No |
 | **Imported** | Reach out to data sources outside DataJoint (instruments, ELNs, external databases) | Yes |
 | **Computed** | Derive their contents entirely from upstream DataJoint tables | Yes |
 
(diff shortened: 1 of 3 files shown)


MilagrosMarin (approved):
LGTM

PR #211, opened 2026-07-17
docs(trace): describe self.upstream as lazily built on first access

MilagrosMarin:
Post-[#1499](https://github.com/datajoint/datajoint-python/pull/1499) (shipped in v2.3.1), `self.upstream` is no longer constructed eagerly before every `make()` — the framework records the key, and `Diagram.trace(self & key)` is built the first time the user accesses `self.upstream` and memoized for the rest of the call. `make()` implementations that never read `self.upstream` pay nothing.

The shipped code (`src/datajoint/autopopulate.py:135-141`, `:673-674`) and the release-notes performance highlight both describe this behavior, but two docs paragraphs still describe the pre-#1499 lifecycle.

## Changes

**`src/reference/specs/trace.md`** — §"Where it's set"
- "constructs `self.upstream = Diagram.trace(self & key)` immediately before invoking" → "records the current key but defers constructing the trace; built on first access, memoized for the rest of the call".
- Preserves the SQL-lazy / fetch-not-cached wording (still accurate).

**`src/reference/specs/autopopulate.md`** — §4.4 `self.upstream` behavior
- Lifecycle bullet: same substitution as above.
- Lazy bullet: adds the "at most once" and "make() calls that never read upstream pay nothing" nuance.

### DIFF
--- a/src/reference/specs/autopopulate.md
+++ b/src/reference/specs/autopopulate.md
@@ -357,8 +357,8 @@ DataJoint 2.3 introduced `self.upstream` (and the underlying [`Diagram.trace`](d
 
 `self.upstream` behavior:
 
-- **Lifecycle:** `self.upstream` is set to `Diagram.trace(self & key)` immediately before `make()` is invoked and cleared afterward — including when `make()` raises. Accessing it outside `make()` raises `DataJointError`.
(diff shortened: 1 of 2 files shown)


dimitri-yatsenko (approved):
LGTM. Accurately describes the lazy `self.upstream` behavior merged in #1499 — key recorded before make(), trace built on first access and memoized, never built if unused, cleared in finally.

PR #212, opened 2026-07-17
docs(whats-new-23): add 2.3.1 cascade fix bullet; refresh releases link

MilagrosMarin:
Two small follow-ups on `whats-new-23.md` after v2.3.1 shipped.

## Changes

**Cascade multi-part-master fix** ([datajoint-python#1484](https://github.com/datajoint/datajoint-python/pull/1484)) — added as the third bullet under "Changes in 2.3.1". The section listed strict_provenance removal, Python 3.14, and the GC fix, but the multi-part-master cascade fix — a silent-integrity path where two Parts of the same Master could leave the Master with rows that should have been deleted — was omitted despite being a v2.3.1 highlight in the release announcement.

**Stale release-notes link** — `[Release Notes (v2.3.0)](.../releases)` in See Also updated to `[Release Notes (v2.3.1)](.../releases/tag/v2.3.1)`.

Docs-only.


dimitri-yatsenko:
The cascade-fix bullet and the `v2.3.1` release-link refresh are both great — please keep those.

One trim: I'd **drop the sentence "Invalid `part_integrity` values now raise `ValueError`."** Two reasons:

1. **It's assumed behavior.** `ValueError` is the expected result for an invalid enum-like argument (and intentionally so — we reserve `DataJointError` for DataJoint-specific logic, so a bad argument value correctly raises the stdlib `ValueError`). It doesn't need a callout.
2. **More importantly, "now" mis-dates it.** `Table.delete()` has raised `ValueError` on invalid `part_integrity` since **2.3.0** — the check shipped in `cabdb744` (an ancestor of `v2.3.0`, `table.py:841-842`). What 2.3.1 (`179c39be`) actually did was replicate the same validation into `Diagram.cascade` and add a `drop()` variant. So as a *What's New in 2.3.1* line it reads as a change that, for `delete()`, isn't one.

Same reasoning underlies the suggestion on #213. Everything else in this PR is good to merge once that sentence is out.

MilagrosMarin:
Thanks — you're right that the `ValueError` line mis-dates the delete() side (the raise was in 2.3.0, not new in 2.3.1) and doesn't add reader value alongside the values table. Dropped in 07f3818. Ready for another look.

src/about/whats-new-23.md
```diff
@@ -105,4 +106,4 @@ It is PostgreSQL-only (raising a clear error on other backends), idempotent at t
 - [Deployment Operations](../reference/specs/deploy-operations.md) — the `dj.deploy` module
 - [Cascade Specification](../reference/specs/cascade.md) — propagation rules shared with `trace`
 - [What's New in 2.2](whats-new-22.md) — Previous release
-- [Release Notes (v2.3.0)](https://github.com/datajoint/datajoint-python/releases) — GitHub changelog
+- [Release Notes (v2.3.1)](https://github.com/datajoint/datajoint-python/releases/tag/v2.3.1) — GitHub changelog
```
dimitri-yatsenko: This should cover all 2.3.* releases.

MilagrosMarin:
Good catch — you're right, the page covers the whole 2.3 line. Pushed 9b4a903: text is now 'Release Notes (2.3.x)' pointing to /releases so it stays valid across 2.3.1, 2.3.2, etc.

dimitri-yatsenko (approved):
LGTM. The `ValueError` callout is dropped (it restated assumed behavior and mis-dated a check that shipped in 2.3.0), and the cascade-fix bullet reads well. Generalizing the link to `2.3.x → /releases` is a good call — stable across the series and avoids pointing at a tag before it's published.

PR #213, opened 2026-07-17
docs(delete): note that invalid part_integrity values raise ValueError

MilagrosMarin:
The invalid-value contract for `part_integrity` — "any value other than `enforce` / `ignore` / `cascade` raises `ValueError`" — was documented only in [`diagram.md:157`](https://github.com/datajoint/datajoint-docs/blob/main/src/reference/specs/diagram.md#L157). Three other spec pages list the three valid values but never mention what happens on an invalid one.

## Changes

- `data-manipulation.md` §4.1 (delete signature): one-line note after the values table.
- `data-manipulation.md` §4.6 (Part.delete): same.
- `master-part.md` §4.2 (Part Integrity Parameter): same.

Note wording mirrors `diagram.md:157` verbatim: *"Invalid values raise `ValueError`."*

Docs-only.

### DIFF
--- a/src/reference/specs/data-manipulation.md
+++ b/src/reference/specs/data-manipulation.md
@@ -320,6 +320,8 @@ def delete(
 | `"ignore"` | Allow deleting parts without masters (breaks integrity) |
 | `"cascade"` | Also delete masters when parts are deleted |
 
+Invalid values raise `ValueError`.
+
 **Returns:** Number of deleted rows from the primary table.
 
 ### 4.2 Cascade Behavior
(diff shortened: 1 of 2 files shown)


dimitri-yatsenko:
Thanks for chasing consistency across the pages — but I'd suggest we **decline this one**, and it's a judgment call about doc value rather than correctness.

Documenting that invalid `part_integrity` values raise `ValueError` restates the assumed behavior. The values table already lists `enforce`/`ignore`/`cascade`, so a reader infers that anything else is invalid, and `ValueError` is exactly the standard-library result a Python caller expects for a bad enum-like argument — that's the whole purpose of `ValueError`. The exception type here is deliberate, too: we reserve `DataJointError` for DataJoint-specific logic, so a plain invalid-argument value correctly falls through to `ValueError`. Given that, spelling it out on three reference pages adds a line without adding information.

Where documenting an exception earns its place is a docstring `Raises:` section (convention — and `table.py`'s `delete()` already has one), or when the behavior is genuinely surprising (a silent default, coercion, or a non-standard exception). None of those apply here.

No concern with the accuracy of the wording — just proposing we keep the reference prose lean and close this. Happy to be overruled if you've seen users actually trip on it.

MilagrosMarin:
Agreed — the values table + stdlib convention already carries this, and a docstring `Raises:` section is the right home when it's needed (which `table.py:delete()` already has). Closing.

PR #215, opened 2026-07-17
docs(diagram): renamed FKs as orange edges

dimitri-yatsenko:
> 🚫 **Do not merge until the v2.3.2 release.** This documents the renamed-FK
> diagram rendering introduced in datajoint/datajoint-python#1508, whose notebooks
> here were re-executed against that (unreleased) branch. Merge only once #1508
> has shipped in **v2.3.2** — and re-run the authoritative full notebook
> re-execution against the *released* package at that time (the #210 pattern).

Paired with **datajoint/datajoint-python#1508** (nx.MultiDiGraph migration, milestone v2.3.2).

## What changed

The diagram notation for a **renamed / aliased foreign key** changed: it used to
render as a small **orange dot** (an integer "alias node" between two tables); it
now renders as a distinctly-colored **orange edge** with a hover tooltip listing
the renamed columns (e.g. `spouse1 ← person_id`). The alias node is gone from the
underlying graph entirely (it is now an `nx.MultiDiGraph` with parallel edges).

### Prose / notation
- `how-to/read-diagrams.ipynb`, `how-to/model-relationships.ipynb`,
  `tutorials/basics/02-schema-design.ipynb` — "orange dots" → "orange edges";
  hover-tooltip wording updated.
- `reference/specs/diagram.md` — new **Edge Styles** row for the renamed-FK orange edge.
- `reference/specs/cascade.md`, `reference/specs/trace.md` — internal-model prose
  updated from the "alias node" mechanism to the MultiDiGraph parallel-edge model.

### New example (shows the semantics are retained)
`read-diagrams.ipynb` gains a **"Renamed primary vs. secondary foreign keys"**
subsection. Orange is layered on top of the usual line styles — it never overrides
them:

| Example | Renamed FK kind | Renders as |
|---|---|---|
| `Person → Marriage` (`spouse1`/`spouse2`) | secondary, multi | dashed orange, thin |
| `Neuron → Synapse` (`presynaptic`/`postsynaptic`) | primary, multi | **solid** orange, thin |
| `Neuron → NeuronProfile` (`subject`) | 1:1 primary | **solid** orange, **thick** |

## Verification
- `read-diagrams` and `model-relationships` notebooks were re-executed against the
  #1508 build (MySQL via the repo's `docker compose` EXECUTE flow). The committed
  SVGs now show `stroke="#ff8800"` on the renamed-FK **edges** (with
  `xlink:title="spouse1 ← person_id"` tooltips) and no alias-node dots; solid/dashed
[...]

PR #216, opened 2026-07-18
docs: interlink schema-design concepts; fix PK/entity-integrity examples

dimitri-yatsenko:
Interlinks the closely-related schema-design concepts — **normalization, primary keys, foreign keys / modeling relationships, entity integrity** — both conceptually and through mutual cross-references, and fixes two misleading examples. Independent of the v2.3.2 code work (no notebook re-execution); separate from datajoint/datajoint-docs#215.

## Content fixes

- **`explanation/entity-integrity.md` — "Too many key attributes".** The example declared `(subject_id, timestamp)` as "Wrong… destroys entity identity" — but that's a *valid* per-subject measurement key. Rewrote it to say so, and replaced the example with genuine over-keying (a **descriptive** attribute — `technician` — placed in the key). Switched to core types (`int32`).
- **`how-to/design-primary-keys.md` — "Keep Keys Small".** Added a genuinely-oversized example (`varchar(200)` where an `int32` suffices — the key is copied into every child table, index, and join) and dropped the incorrect `# Max 65,535 sessions` comment on an `int32` field (65,535 is `int16`'s range).

## Interlinking (the family was barely connected; `normalization.md` had no See-also at all)

- **`design-primary-keys.md` ↔ `entity-integrity.md`** — mutual links; the how-to now states the primary key is *how a table enforces entity integrity*.
- **`normalization.md`** — new "See also" → entity-integrity, design-primary-keys, model-relationships, relational-workflow-model.
- **`entity-integrity.md`** "See also" → design-primary-keys, model-relationships.
- **`relational-workflow-model.md`** "See also" → the schema-design how-tos.

## Notes

- `model-relationships.ipynb` is intentionally untouched here (it's owned by #215); the one remaining backlink (model-relationships → normalization) can be added there or in a follow-up.
- Verified all added relative links resolve to existing files.

### DIFF
--- a/src/explanation/entity-integrity.md
+++ b/src/explanation/entity-integrity.md
@@ -186,6 +186,11 @@ key. Neither attribute alone would identify a recording across the table.
 
 ## Foreign keys
 
+Referential integrity is impossible without entity integrity: you must be able
+to identify unique entities before you can define valid relationships between
+them. A foreign key builds on the parent's primary key, so a well-designed key
+is the prerequisite for every relationship that references it.
+
 A foreign key serves two distinct purposes in DataJoint:
 
 1. **Referential integrity.** The referenced entity must exist.
--- a/src/explanation/entity-integrity.md
+++ b/src/explanation/entity-integrity.md
@@ -412,16 +412,26 @@ class Trial(dj.Manual):
 
 ### Too many key attributes
 
+`(subject_id, timestamp)` is a *correct* key for a per-subject measurement:
+each reading is its own entity, identified by who was measured and when.
+Over-keying is the opposite mistake — putting a **descriptive** attribute into
+the key, so a single entity fragments into several records:
+
 ```python
-# Wrong: timestamp makes every row unique, losing entity semantics
+# Wrong: `technician` describes who took the reading; it doesn't identify it —
+# the same measurement recorded under two technicians becomes two entities
 class Measurement(dj.Manual):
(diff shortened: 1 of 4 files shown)


MilagrosMarin (approved):
Thanks — the interlinking is a real improvement, and both content fixes are correct: the `(subject_id, timestamp)` example was misleading (that's a valid per-subject measurement key), and the `# Max 65,535 sessions` comment on `int32` was wrong (that's int16's range). Verified all internal links resolve and CI (spelling + linkcheck) passes. New three-lens framing in `normalization.md` reads well.

Two small observations, not blockers:

1. **`design-primary-keys.md` drops "Use Fixed-Width for Joins" and "Avoid Computed Values"** without a note in the PR body. Both deletions are defensible — the char-vs-varchar guidance is arguably dated (modern InnoDB handles VARCHAR joins fine), and "avoid computed values" would conflict with DJ 2.x's hash-addressed storage as a first-class concept. Just calling it out in case one of them was meant to survive in some form.

2. **The e-commerce example in `normalization.md` uses `dj.Manual` for `Payment`, `Shipment`, `Delivery`.** Pedagogically these would more naturally be `dj.Imported` (payment processor / carrier API) or auto-populated by workflow events; `dj.Manual` glosses over the ingestion mechanism. Minor — the point being made (workflow > 3NF) still lands cleanly.

Approving.

Issue #225, opened 2026-07-18
Docs: represent the DataJoint book's core principles (coverage review)


dimitri-yatsenko:
Umbrella tracking a **book → docs coverage review**: an audit of the core principles taught in the [DataJoint book](https://github.com/datajoint/datajoint-book) against the published documentation. Across ~37 principles in three clusters (integrity/keys, normalization, computation/reproducibility), the docs are solid overall — and stronger than the book on the `make()` reproducibility contract — with the gaps below.

### New pages / major expansions
- [ ] #217 — Integrity-types explanation page (the seven integrity types + rules→constraints map)
- [ ] #218 — Referential-integrity / foreign-keys explanation page (FK = runtime logical constraint, not a pointer; the five effects incl. auto-index)
- [ ] #219 — Expand `normalization.md` to the three-lens treatment (+ the Order example that shows workflow normalization is stricter than 3NF)
- [ ] #220 — Transactions / ACID explanation page

### Targeted additions & one fix
- [ ] #221 — `insert-data` how-to: insertion-discipline / reproducibility guardrails _(in progress)_
- [ ] #222 — master-part spec: "a dependency on the master implies dependency on all its parts"
- [ ] #223 — An "update data" how-to (updates are rare; delete + reinsert + recompute)
- [ ] #224 — Fix: `autopopulate.md` §5.5 shows a transaction inside `make()`, contradicting `computation-model.md`

### Already addressed while this review ran
The audit baselined `main`, so these showed as gaps there but are closed on in-flight branches:
- "Referential integrity is impossible without entity integrity" → `entity-integrity.md` (#216)
- Natural-key "real-world use" test in the how-to → `design-primary-keys.md` (#216)
- Concept interlinking (normalization was a link-island; PK ↔ entity-integrity mutual) (#216)
- Primary-key guideline corrections — dropped the fixed-width "joins faster" myth, "avoid dates", "avoid computed values"; added the floating-point-key warning (#216)
- Delete → garbage-collection lifecycle link → `delete-data.md`

### Where the docs already win (no action)
The `make()` reproducibility contract (formalized beyond the book), no-auto-increment rationale, the three questions of entity integrity, schema dimensions & lineage, cardinality-from-FK-placement, tripartite `make()`, and the mouse/cage normalization example.

dimitri-yatsenko:
**Progress — PRs opened for all of P1–P8:**

[...]

Issue #224, opened 2026-07-18
Fix contradiction: autopopulate.md §5.5 shows a transaction inside make()


dimitri-yatsenko:
`reference/specs/autopopulate.md` **§5.5 "Manual Transaction Control"** shows opening a transaction *inside* `make()`:

```python
def make(self, key):
    with dj.conn().transaction:
        self.insert1(...)
        self.Part.insert(parts)
```

This **directly contradicts** `explanation/computation-model.md` ("Don't Open a Transaction Inside make()"), which states that `populate()` already holds a transaction for the call, so starting a nested one **raises an error** (DataJoint doesn't support nested transactions).

One of the two is wrong. **Fix §5.5** — remove/replace the in-`make()` transaction example and align it with `computation-model.md` (a `make()` is already atomic; no explicit transaction is needed or allowed inside it).

_From an internal datajoint-book → docs coverage review (priority P8)._

dimitri-yatsenko:
Addressed by #226, which fixed `reference/specs/autopopulate.md` §5.5 — removed the transaction-inside-`make()` example that contradicted `computation-model.md` (`make()` is already transactional; a nested one raises). Closing.

Issue #223, opened 2026-07-18
Add an 'update data' how-to


dimitri-yatsenko:
The book's updates chapter (`40-operations/030-updates.ipynb`) frames updates as **rare and deliberate** — prefer delete + reinsert + recompute; `update1()` is a surgical correction, not a normal workflow operation. The docs have **no how-to counterpart**; `update1()` / when-to-update guidance is scattered across explanation pages and specs.

**Proposal:** add `how-to/update-data.md` (or a section in an existing how-to) covering when updates are appropriate, `update1()` usage, and the delete-reinsert-recompute pattern for correcting computed results.

_From an internal datajoint-book → docs coverage review (priority P7)._

dimitri-yatsenko:
Addressed by #226, which added the `how-to/update-data.md` how-to — updates are rare and surgical, prefer delete + reinsert (+ recompute), `update1()` for single-row corrections. Closing.

Issue #222, opened 2026-07-18
master-part spec: state that a dependency on the master implies its parts


dimitri-yatsenko:
The book (`30-design/053-master-part.ipynb`) teaches that **a dependency on a master is transitively a dependency on all of its parts** — referencing the master guarantees every part row exists. This is a design consequence downstream tables rely on.

`reference/specs/master-part.md` covers structure, delete, and query semantics but does **not** state this transitive-completeness guarantee.

**Proposal:** add a subsection to `reference/specs/master-part.md` stating the guarantee, and a short note in `how-to/master-part.ipynb`.

_From an internal datajoint-book → docs coverage review (priority P6)._

dimitri-yatsenko:
Addressed by #226, which added the transitive-completeness subsection to `reference/specs/master-part.md` — a dependency on the master is a dependency on all of its parts. Closing.

PR #226, opened 2026-07-18
docs(data-ops): insertion discipline, update-data how-to, master-part & make-transaction fixes

dimitri-yatsenko:
Data-operation how-to and spec improvements from the [book → docs coverage review](https://github.com/datajoint/datajoint-docs/issues/225). Closes **#221, #222, #223, #224**, and adds the delete → garbage-collection cross-reference.

### What's here
- **`insert-data.md` — "What Not to Insert"** (#221): the how-to now warns that the examples target Manual tables and you must not hand-insert into **Computed/Imported** (rows come from `make()`/`populate()`), **Lookup** rows come from `contents` in the definition, insert **parents before children**, and never open a transaction inside `make()`.
- **`update-data.md` — new how-to** (#223): updates are rare and surgical; prefer **delete + reinsert (+ recompute)**; `update1()` for single-row corrections; primary keys and computed results can't be updated.
- **`master-part.md` spec** (#222): adds **transitive completeness** — a dependency on the master is a dependency on all of its parts.
- **`autopopulate.md` spec §5.5 fix** (#224): it showed a transaction opened *inside* `make()`, contradicting `computation-model.md`; `make()` is already transactional (a nested one raises). Replaced with correct guidance.
- **`delete-data.md`**: "Reclaiming Object Storage" note + link to the garbage-collection how-to (deletes leave in-store objects; GC reclaims them).

Draft for review on the docs server. No dependency on the v2.3.x code work.

### DIFF
--- a/mkdocs.yaml
+++ b/mkdocs.yaml
@@ -78,6 +78,7 @@ nav:
           - Query Data: how-to/query-data.md
           - Fetch Results: how-to/fetch-results.md
           - Delete Data: how-to/delete-data.md
+          - Update Data: how-to/update-data.md
       - Computation:
           - Run Computations: how-to/run-computations.md
           - Distributed Computing: how-to/distributed-computing.md
--- a/src/how-to/delete-data.md
+++ b/src/how-to/delete-data.md
@@ -237,4 +237,5 @@ after large deletes.
 - [Master-Part Tables](master-part.ipynb) — Compositional data patterns
 - [Model Relationships](model-relationships.ipynb) — Foreign key patterns
 - [Insert Data](insert-data.md) — Adding data to tables
+- [Update Data](update-data.md) — Correcting or replacing existing rows
 - [Run Computations](run-computations.md) — Recomputing after changes
--- a/src/how-to/delete-data.md
+++ b/src/how-to/delete-data.md
@@ -221,8 +221,18 @@ cascade
 
 For simple single-table deletes, `(Table & restriction).delete()` remains the simplest approach. The diagram API is for when you need more visibility before executing.
 
+## Reclaiming Object Storage
+
+Deleting rows removes their database records, but for tables with in-store types
+(`<blob@>`, `<attach@>`, `<object@>`, `<npy@>`) the stored objects remain — by
+design (hash-addressed storage is deduplicated, and immediate deletion is unsafe
+under concurrency). Those orphaned objects are reclaimed separately by
+[garbage collection](garbage-collection.md); run it periodically to free storage
+after large deletes.
+
 ## See Also
 
+- [Clean Up Object Storage](garbage-collection.md) — reclaim stored objects orphaned by deletes
 - [Diagram Specification](../reference/specs/diagram.md) — Full reference for diagram operations
 - [Master-Part Tables](master-part.ipynb) — Compositional data patterns
 - [Model Relationships](model-relationships.ipynb) — Foreign key patterns
--- a/src/how-to/delete-data.md
+++ b/src/how-to/delete-data.md
@@ -87,7 +87,7 @@ Deletes are atomic—all cascading deletes succeed or none do:
 Within an existing transaction:
 
 ```python
-with dj.conn().transaction:
+with schema.connection.transaction:
     (Table1 & key1).delete(transaction=False)
     (Table2 & key2).delete(transaction=False)
     Table3.insert(rows)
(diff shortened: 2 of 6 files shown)


MilagrosMarin (changes requested):
Thanks — most of this is solid and verified:

- §5.5 make-transaction fix (#224) is exactly right; `connection.py:523` explicitly raises "Nested connections are not supported" if `start_transaction()` fires during an active transaction, and `_populate_one` (`autopopulate.py:664`) opens one before `make()`. So the old §5.5 example would raise.
- `insert-data.md` "What Not to Insert" claims all accurate (Computed/Imported reject direct inserts via `_allow_insert`; Lookup rows via `contents`; parents-before-children; no-transactions-in-make()).
- `update-data.md` reads cleanly; delete+reinsert vs `update1()` distinction is correct, and the primary-key immutability + no-recompute caveats are right.
- `delete-data.md` GC cross-reference lands accurately.
- All CI green; all internal links resolve.

**One blocker on the `master-part.md` §2.1 addition:**

The new sentence — *"Downstream tables therefore reference the master alone and can rely on the whole composite being present — they do not (and cannot) depend on the parts individually."* — the `(and cannot)` is too strong and contradicts standard DataJoint practice:

- The framework's own tests use downstream FK references to Part tables:
  - `tests/schema_simple.py:126` → `-> B.C`
  - `tests/schema_simple.py:148` → `-> E.H`
  - `tests/integration/test_schema.py:251` → `-> Schema_A.Subject`
  - `tests/integration/test_cascade_delete.py:326` → `-> Master.PartA`
  - `tests/integration/test_cascade_delete.py:373` → `-> Master.PartA.proj(...)` (renamed-FK case)
- The Elements ecosystem uses the same pattern widely (e.g., `-> Segmentation.Mask` in `element-calcium-imaging`).

If the intent is *"you don't need to reference the parts individually because a master reference reaches the whole composite,"* that's a useful design guideline — but "cannot" reads as a hard rule and contradicts a legal, common, and framework-tested pattern. Would you soften along the lines of *"typically reference the master alone; a foreign key to a specific Part is legal but often unnecessary when the composite is what matters"* or equivalent?

**Related, smaller clarification on the same paragraph:**

*"Because the master and its parts are inserted and deleted as one unit"* — deletion is FK-CASCADE enforced, but insertion atomicity is only enforced when `make()` inserts both (Computed/Imported via `populate()`'s wrapping transaction). For Manual master-part, users can insert `Master.insert1(...)` without any Part rows and the RDBMS accepts it; atomic insertion is convention (`with dj.conn().transaction: Master.insert1(...); Master.Part.insert(...)`), not enforcement. Reads as a mechanical guarantee today — worth a brief clarification either way.

Happy to re-review once §2.1 is adjusted.

dimitri-yatsenko:
Thanks @MilagrosMarin — addressed both points in \`master-part.md\` §2.1 (68bf9d9):

- **Blocker:** dropped the "(and cannot) depend on the parts individually" framing. The paragraph now states downstream tables *typically* reference the master alone, and explicitly notes a foreign key to a specific Part is legal and common in practice — just often unnecessary when the composite as a whole is what matters.
- **Clarification:** insertion atomicity is now distinguished by tier — enforced for Computed/Imported (`populate()` wraps each `make()` in a transaction), convention for Manual master-part sets (`with dj.conn().transaction:`). Deletion cascade noted as holding in either case.

Ready for re-review.

MilagrosMarin (dismissed):
Thanks — both changes read exactly right. The new "typically … a foreign key to a specific part is still legal and common in practice, but often unnecessary" framing acknowledges the pattern DJ's own tests use, and the tier-split on insertion atomicity (make()-enforced for Computed/Imported vs convention for Manual) correctly separates the two cases while keeping the deletion-cascade universality intact. Approving.

dimitri-yatsenko:
Additional fix pushed (aa4a1bc8): corrected the Manual-tier definition in `insert-data.md` — "the tier whose rows are entered by hand" → "the tier whose rows are inserted directly, from outside the DataJoint pipeline (whether entered by hand through a form or GUI, or loaded by an automated ingestion tool)". A broader tier-definition consistency sweep across the tutorials and quick-reference tables is in #230.

MilagrosMarin (dismissed):
LGTM

dimitri-yatsenko:
Converted the transaction examples/mentions in this PR's pages to the instance-based idiom (`schema.connection.transaction`), matching the docs-wide sweep in #231, so the pages are internally consistent. (992e44c7)

MilagrosMarin (dismissed):

MilagrosMarin (approved):

Issue #220, opened 2026-07-18
Add a Transactions / ACID explanation page


dimitri-yatsenko:
The book devotes a full chapter (`40-operations/040-transactions.ipynb`) to **data consistency, ACID, CAP, and a bank-transfer failure demonstration**. The docs cover transactions only *operationally* — a `delete(transaction=False)` flag and the automatic `make()` transaction — with no conceptual page.

**Proposal:** add `explanation/transactions.md` porting the ACID / consistency framing (what a transaction guarantees, atomic all-or-nothing, the bank-transfer example), and link it from the delete/insert how-tos and the computation model.

_From an internal datajoint-book → docs coverage review (priority P4)._

dimitri-yatsenko:
Addressed by #227, which added `explanation/transactions.md` — atomic all-or-nothing commits, ACID grounded in DataJoint, delete atomicity, and why `make()` must not open its own transaction. Closing.

Issue #218, opened 2026-07-18
Add a referential-integrity / foreign-keys explanation page


dimitri-yatsenko:
The book has a dedicated **Foreign Keys** chapter (`30-design/030-foreign-keys.md`); the docs split FK material between `explanation/entity-integrity.md` and the `how-to/model-relationships.ipynb` how-to, with **no dedicated referential-integrity explanation page** — an asymmetry with the book's entity-key / foreign-key structure.

Concepts currently missing:
- **A foreign key is a runtime logical constraint, not a physical pointer** — an insert triggers a search of the parent to verify a matching key exists; it is not a disk link. (The HDF5-pointer contrast is worth keeping.)
- **The five effects of a foreign key** — schema embedding, insert restriction on the child, delete restriction on the parent (cascade), update restriction on keys, and **automatic index creation**. They're never enumerated together, and the auto-index effect is entirely absent from the docs.

**Proposal:** add `explanation/referential-integrity.md` covering both, plus the "referential integrity is impossible without entity integrity" dependency (now added to `entity-integrity.md` in PR #216 — link it).

_From an internal datajoint-book → docs coverage review (priority P2)._

dimitri-yatsenko:
Addressed by #227, which added `explanation/referential-integrity.md` — foreign key as a runtime logical constraint (not a pointer), the five effects of an FK including automatic index creation, and the "referential integrity is impossible without entity integrity" dependency. Closing.

Issue #217, opened 2026-07-18
Add an integrity-types explanation page (the seven integrity types)


dimitri-yatsenko:
The DataJoint book (`20-concepts/04-integrity.md`) organizes the whole data model around **seven integrity types** — domain, completeness, entity, referential, compositional, consistency, workflow — each mapped to a concrete table feature, plus a real-world-rule → database-constraint mapping. It's the book's single most structuring idea.

The docs cover entity / referential / compositional integrity only in pieces (spread across `explanation/entity-integrity.md` and `reference/specs/master-part.md`), and **domain, completeness, consistency, and workflow integrity are never named as integrity types**. There is no page that enumerates the seven.

**Proposal:** add `explanation/data-integrity.md` enumerating all seven integrity types with the rules→constraints table, and cross-link the existing entity/referential/compositional material into it (and from `explanation/index.md`).

_From an internal datajoint-book → docs coverage review (priority P1)._

PR #227, opened 2026-07-18
docs: add Data Integrity, Referential Integrity, and Transactions explanation pages

dimitri-yatsenko:
Three new **explanation** pages from the [book → docs coverage review](https://github.com/datajoint/datajoint-docs/issues/225), filling the biggest structural gaps in the Data Model concept cluster. Closes **#217, #218, #220**.

- **`explanation/data-integrity.md`** (#217) — the book's organizing idea: the **seven integrity types** (domain, completeness, entity, referential, compositional, consistency, workflow), each mapped to the DataJoint feature that enforces it, plus the real-world-rule → constraint decomposition.
- **`explanation/referential-integrity.md`** (#218) — the FK explanation the docs lacked: a foreign key is a **runtime logical constraint, not a pointer**; the **five effects** of an FK (incl. automatic index creation); the referential-integrity + workflow-DAG dual role. Opens with "referential integrity is impossible without entity integrity."
- **`explanation/transactions.md`** (#220) — atomic all-or-nothing commits, ACID grounded in DataJoint, delete atomicity, and why `make()` (already wrapped in a transaction by `populate()`) must not open its own.

All three are in the Data Model nav and cross-link the existing pages. Conventions verified: core types only, "diagrams" not ERDs, no "provenance", object store framed as internal, MySQL/PostgreSQL as peers.

Draft for review on the docs server.

### DIFF
--- a/mkdocs.yaml
+++ b/mkdocs.yaml
@@ -13,8 +13,11 @@ nav:
           - FAQ: explanation/faq.md
       - Data Model:
           - Relational Workflow Model: explanation/relational-workflow-model.md
+          - Data Integrity: explanation/data-integrity.md
           - Entity Integrity: explanation/entity-integrity.md
+          - Referential Integrity: explanation/referential-integrity.md
           - Normalization: explanation/normalization.md
+          - Transactions: explanation/transactions.md
           - Computation Model: explanation/computation-model.md
           - Fan-Out Ingestion: explanation/fan-out-ingestion.md
           - Schema as a Workflow Specification: explanation/schema-as-workflow-specification.md
--- a/src/about/whats-new-22.md
+++ b/src/about/whats-new-22.md
@@ -266,9 +266,9 @@ When `cascade()` or `restrict()` propagates a restriction from a parent to a chi
 
 | Rule | Condition | Child restriction |
 |------|-----------|-------------------|
-| **Direct copy** | Non-aliased FK, restriction attributes are a subset of child's primary key | Restriction copied directly |
-| **Aliased projection** | FK uses attribute renaming (e.g., `subject_id` → `animal_id`) | Parent projected with attribute mapping |
-| **Full projection** | Non-aliased FK, restriction uses attributes not in child's primary key | Parent projected (all attributes) as restriction |
+| **Direct copy** | Non-renamed FK, restriction attributes are a subset of child's primary key | Restriction copied directly |
+| **Renamed projection** | FK uses attribute renaming (e.g., `subject_id` → `animal_id`) | Parent projected with attribute mapping |
+| **Full projection** | Non-renamed FK, restriction uses attributes not in child's primary key | Parent projected (all attributes) as restriction |
 
 When a child has multiple restricted ancestors, convergence depends on the mode: `cascade()` uses OR (any path marks a row for deletion), `restrict()` uses AND (all conditions must match).
 
--- a/src/explanation/data-integrity.md
+++ b/src/explanation/data-integrity.md
@@ -118,9 +118,12 @@ DataJoint declares references with the foreign-key arrow `->`, which inherits
 the referenced table's primary key into the dependent table and enforces that
 the referenced row exists before a dependent row can be inserted. Removal is
 symmetric: deleting an entity **cascades** to everything that depends on it,
-so the database never enters a state with orphaned references — including
-objects held in integrated object storage, which are removed together with
-the rows that own them. See [Referential Integrity](referential-integrity.md).
+so the database never enters a state with orphaned references. This is a
+row-level guarantee: objects held in integrated object storage are *not* removed
+synchronously with their rows (hash-addressed storage is deduplicated, and
+immediate deletion is unsafe under concurrency) — they are reclaimed
+asynchronously by garbage collection. See
+[Referential Integrity](referential-integrity.md).
 
 ### 5. Compositional integrity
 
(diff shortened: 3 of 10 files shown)


MilagrosMarin (changes requested):
Thanks — three genuinely useful new explanation pages, and the "aliased" → "renamed" terminology sweep matches `definition-syntax.md`'s `[rename]` grammar. `referential-integrity.md` is particularly well-crafted (the FK-as-runtime-constraint framing, the five-effects breakdown, the modifier/placement discussion). Most claims verify cleanly:

- Nested transactions raise — `connection.py:523` explicitly raises `DataJointError("Nested connections are not supported.")`.
- `populate()` wraps `make()` in a transaction — `autopopulate.py:664` `start_transaction()` before make().
- `delete()` transactional by default — `table.py:1016-1030` `transaction: bool = True`.
- All CI green (spelling + linkcheck + check-versions).

**Two factual issues on the new pages worth resolving before merge:**

### 1. `transactions.md` — "serializable-style isolation" is not what DataJoint uses

The page says *"using serializable-style isolation for operations that must be atomic."* Actual code:

- **MySQL** (`adapters/mysql.py:793`) → `START TRANSACTION WITH CONSISTENT SNAPSHOT` — REPEATABLE READ with a consistent-view snapshot (InnoDB default).
- **PostgreSQL** (`adapters/postgres.py:1006`) → plain `BEGIN` — READ COMMITTED (Postgres default).

Neither is SERIALIZABLE, and the two backends actually use *different* default isolation levels — not the same one. "Serializable-style" is a specific term readers familiar with isolation levels will read literally. Suggest something like *"using the backend's default isolation, which provides ACID atomicity and consistency for the grouped statements"* or *"MySQL uses REPEATABLE READ (with a consistent snapshot), PostgreSQL uses READ COMMITTED"* if you want to be explicit.

### 2. `data-integrity.md` §4 — "objects removed together with the rows" contradicts `delete-data.md`

Current text: *"deleting an entity cascades to everything that depends on it, so the database never enters a state with orphaned references — including objects held in integrated object storage, which are removed together with the rows that own them."*

This contradicts what `delete-data.md` (in #226) explicitly states: *"Deleting rows removes their database records, but for tables with in-store types (`<blob@>`, `<attach@>`, `<object@>`, `<npy@>`) the stored objects remain — by design (hash-addressed storage is deduplicated, and immediate deletion is unsafe under concurrency). Those orphaned objects are reclaimed separately by garbage collection."*

The DB-level referential-integrity claim is fine (no dangling FK rows), but the parenthetical about object storage being "removed together" is not accurate — external-store objects persist until `dj.gc.GarbageCollector` reclaims them. Suggest dropping or reframing to *"referential integrity is maintained at the row level; object-storage cleanup is asynchronous — see garbage collection"*.

**Two smaller points, non-blocking but worth a look:**

### 3. Master-part "single transaction" claim — same nuance as #226

Both `data-integrity.md` §5 ("Insertion and deletion of a master and its parts occur as a single transaction") and `transactions.md`'s master-part example read as a mechanical guarantee. Enforced for `make()` (Computed/Imported), but for Manual master-part it's convention (`with dj.conn().transaction:`), not RDBMS-enforced — a user can insert a Manual master without any parts and the DB accepts it. Consistent with the discussion happening on #226; same softening would apply here.

### 4. `referential-integrity.md` "automatic index creation" — verify for PostgreSQL

*"The database automatically creates a secondary index on the child's foreign-key columns."* True for MySQL InnoDB (it auto-indexes FK columns). **PostgreSQL does not** auto-index FK columns by default. If DataJoint's `declare.py` explicitly emits `CREATE INDEX` for FK columns on the Postgres side, the claim is accurate but should say *"DataJoint creates"* rather than *"the database creates"*. I saw index-creation plumbing in `declare.py` but didn't fully trace whether FK columns specifically get one on Postgres — worth checking.

Happy to re-review once these are adjusted.

dimitri-yatsenko:
Thanks @MilagrosMarin — all four points addressed (1f969fd):

**Blocking**
1. `transactions.md` — dropped "serializable-style isolation". Now states the actual defaults: MySQL opens with a consistent snapshot (REPEATABLE READ), PostgreSQL uses READ COMMITTED. Confirmed against `adapters/mysql.py:793` (`START TRANSACTION WITH CONSISTENT SNAPSHOT`) and `adapters/postgres.py:1006` (plain `BEGIN`).
2. `data-integrity.md` §4 — reframed to a **row-level** guarantee; in-store objects are reclaimed asynchronously by garbage collection, not removed with their rows. Now consistent with `delete-data.md` in #226.

**Non-blocking (also fixed)**
3. `data-integrity.md` §5 — master-part insertion atomicity now distinguished: `make()`-enforced for Computed/Imported, convention for Manual. (The `transactions.md` example already used conditional "inserting them as one transaction guarantees…" framing, so it was left as-is.)
4. `referential-integrity.md` §5 — verified in source: DataJoint emits an explicit FK-column index only for the **unique** case (`declare.py:304-307`); otherwise it relies on the backend. MySQL/InnoDB auto-indexes FK columns, **PostgreSQL does not** — so the section now says MySQL auto-creates it and advises an explicit `index(...)` on PostgreSQL. Also renamed the heading to "Indexing the foreign-key columns" (no inbound anchor links).

Ready for re-review.

MilagrosMarin (dismissed):
Thanks — all four addressed cleanly. Isolation-level wording now matches the actual code paths (MySQL consistent-snapshot / PG READ COMMITTED), the §4 rewrite gives referential integrity its correct row-level scope while pointing to GC for the async object cleanup, master-part §5 mirrors the PR #226 tier-split, and the FK-index section now correctly separates MySQL/InnoDB auto-indexing from the PostgreSQL declare-your-own case. Nice heading rename too. Approving.

dimitri-yatsenko:
Converted `transactions.md` to the instance-based idiom (`schema.connection.transaction`; `self.connection.transaction` inside `make()`) and added a "Which connection the transaction uses" note covering the shared-connection scope. Matches the docs-wide sweep #231. (ed8ee29c)

MilagrosMarin (approved):

Issue #205, opened 2026-07-15
Add a complete, executed worked example of the tripartite make()


dimitri-yatsenko:
## Summary

Add a **complete, executed worked example** of the three-part (tripartite) `make` — `make_fetch` / `make_compute` / `make_insert` — as a tutorial. Today the pattern is documented in prose (Computation Model, [`autopopulate.md` §4.5](https://docs.datajoint.com/reference/specs/autopopulate/)) but there is no end-to-end, runnable example.

## Why

- **The calling convention is a footgun.** The framework invokes `make_compute(key, *fetched_data)` and `make_insert(key, *computed_result)` (the returned tuples are unpacked; `make_insert` receives only the computed result, not `fetched`). This is easy to get wrong — the docs themselves shipped an incorrect example recently. A copy-pasteable, correct template is the best antidote.
- **The semantics are subtle.** `make_fetch` runs outside the transaction and is **re-run and hash-verified** inside it (so it must be bitwise-reproducible); `make_compute` runs once and is never re-verified (so it **may be stochastic**); `make_insert` runs inside the transaction. These boundaries are hard to grasp abstractly and land better as concrete, running code.
- **Real need.** It's the recommended pattern for long-running computations (minutes–hours) that can't hold a transaction open — a common ML / signal-processing case.

## What it should show

- All three phases with the **correct signatures** and tuple return/unpack convention.
- The re-fetch/verify behavior, ideally including the **failure mode**: inputs change mid-compute → rollback, no insert.
- A `make_compute` that is legitimately non-deterministic (e.g. a seeded vs. unseeded stochastic step), to make "reproducibility = tracked derivation, not bitwise determinism" concrete.
- How `populate(make_kwargs=...)` forwards to `make_fetch`.

## Recommended form

- **Executed, not prose** — a tutorial notebook run in CI (or a doctest), so it stays correct as the API evolves. A non-executed snippet would drift out of sync again (exactly the failure we just fixed).
- **Placement:** Tutorials → Advanced (e.g. "Long computations with the three-part make"), or How-To → Computation. Pair with, and cross-link, [`autopopulate.md` §4.5](https://docs.datajoint.com/reference/specs/autopopulate/) and the [Computation Model](https://docs.datajoint.com/explanation/computation-model/).

## Notes

- Scope/reach is modest (most pipelines use plain `make()`; tripartite is advanced), but the value for its audience is high and the reliability win — a tested example that can't drift — is the main argument.
[...]

PR #228, opened 2026-07-18
docs: add executed three-part make() worked example (#205)

dimitri-yatsenko:
Closes **#205** — adds a complete, **executed** worked example of the tripartite `make()`.

### What's here
- **`tutorials/advanced/three-part-make.ipynb`** (new, executed): a self-contained example of the three-part `make()` — a `Recording` source table and a `RecordingStats` computed table with a `Window` part. It shows the phase split: `make_fetch` reads the declared input (outside the transaction, re-fetched and hash-verified before insert), `make_compute` runs the slow work with no transaction held, and `make_insert` writes `self` + its part atomically. Core types only (`int32`/`float64`/`<blob>`), no data downloads (signal generated inline with a seeded RNG). **Executed against released `datajoint==2.3.1`** — outputs are baked in, 0 errors.
- **`run-computations.md` fix**: the existing tripartite example had `make_compute(self, key, fetched)` / `make_insert(self, key, fetched, computed)`, but the runtime unpacks each phase's returned tuple into the next — so the correct signatures are `make_compute(self, key, data)` and `make_insert(self, key, result)` (as in `computation-model.md` and the spec). The old form would raise `TypeError` at populate time.
- Added to the Advanced tutorials nav.

Independent of the v2.3.2 code work — runs against released 2.3.1.

### DIFF
--- a/mkdocs.yaml
+++ b/mkdocs.yaml
@@ -55,6 +55,7 @@ nav:
           - Distributed Computing: tutorials/advanced/distributed.ipynb
           - Custom Codecs: tutorials/advanced/custom-codecs.ipynb
           - Instances: tutorials/advanced/instances.ipynb
+          - Three-Part make(): tutorials/advanced/three-part-make.ipynb
   - How-To:
       - how-to/index.md
       - Setup:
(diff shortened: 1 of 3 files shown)


MilagrosMarin (approved):
LGTM

PR #230, opened 2026-07-20
docs: consistent table-tier definitions (Manual = direct insert, not 'by hand')

dimitri-yatsenko:
## What

Makes the **table-tier definitions consistent** across the docs. The explanation and reference-spec layers already define the tiers correctly (`explanation/relational-workflow-model.md`, `reference/specs/autopopulate.md`); the tutorials and two quick-reference tables had drifted into describing **Manual** tables as "user-entered" / "entered by hand," which wrongly excludes automated ingestion.

## Why

The defining property of a **Manual** table is that its rows are **inserted directly, from outside the DataJoint pipeline** — whether typed in by a person (form/GUI) *or* loaded by an automated ingestion tool/instrument. The contrast that matters is *direct insertion* vs. *production by `make()`* (Imported/Computed), not human vs. machine authorship. Several tutorial tier-tables stated the human-entry framing as the definition, and a few narrowed **Imported** to "files" rather than any external source.

## Changes

- **Manual** → "data inserted directly from outside the pipeline (people/forms *or* ingestion scripts/instruments)".
- **Imported** → "populated by `make()` from an external source (files, instruments, databases)" (no longer "files" only).
- **Computed** → "populated by `make()` from other tables".

Files: `how-to/define-tables.md`, `reference/specs/table-declaration.md`, and the tier tables/legends in the `sql-comparison`, `calcium-imaging`, `blob-detection`, and `basics/02-schema-design` tutorial notebooks. Notebook edits are text-only (no re-execution; outputs untouched).

Scope note: intentionally left the borderline intro-tutorial phrasings (`01-first-pipeline` "you enter data directly," `comparison-to-provenance-systems.md` "manual data-entry point") as-is; they keep the correct "directly" framing and rewording adds churn without changing meaning.


MilagrosMarin (approved):
LGTM

PR #231, opened 2026-07-20
docs: instance-based transaction idiom (schema.connection.transaction)

dimitri-yatsenko:
## What

Global sweep of transaction examples from the global-singleton `dj.conn().transaction` to the **instance-based** form. 17 occurrences across 9 files.

- **Module-level examples → `schema.connection.transaction`** — the `schema` object is defined at module top in every pipeline (the `schema = dj.Schema(...)` / `@schema` convention), so it is in scope, and a transaction usually spans several tables in that schema.
- **In-method examples → `self.connection.transaction`** — where `self` is in scope (`explanation/computation-model.md` WRONG-nesting example; `reference/specs/autopopulate.md`).

## Why

`dj.conn()` returns the singleton `Connection`, and a `Schema` binds every table to that same object (`schemas.py:285` `table._connection = schema.connection`), so at runtime the two are identical **in the default single-connection case**. But `schema.connection.transaction`:

1. names the connection at the natural altitude (the schema owns the tables the transaction spans), rather than through an arbitrary table or a process-global accessor;
2. stays correct in genuinely instance-based setups — if tables live on a non-default connection (a second `dj.Instance`, an explicit `connection=` on the schema), `dj.conn()` returns the *default* singleton, a different object, and would wrap the wrong connection.

## Scope

Mechanical idiom change only. Other `dj.conn()` uses (the accessor itself, `.query(...)`, `.connect()`, thread-safe-mode references) are intentionally left untouched. Notebook edits are text-only (no re-execution). 

Note: the new explanation pages in #226/#227 (`transactions.md`, `master-part.md`, `insert-data.md`, `delete-data.md`) are not on `main` yet, so they are **not** covered here — they still use `dj.conn().transaction` and can be swept once they merge (or in those PRs). A one-line shared-connection caveat belongs on `transactions.md` (in #227); happy to add it there.

### DIFF
--- a/src/explanation/computation-model.md
+++ b/src/explanation/computation-model.md
@@ -396,7 +396,7 @@ raises an error:
 ```python
 def make(self, key):
     # WRONG — a transaction is already in progress, so this raises an error
-    with dj.conn().transaction:
+    with self.connection.transaction:
         self.insert(compute_multiple_results(key))
 ```
 
--- a/src/how-to/delete-data.md
+++ b/src/how-to/delete-data.md
@@ -87,7 +87,7 @@ Deletes are atomic—all cascading deletes succeed or none do:
 Within an existing transaction:
 
 ```python
-with dj.conn().transaction:
+with schema.connection.transaction:
     (Table1 & key1).delete(transaction=False)
     (Table2 & key2).delete(transaction=False)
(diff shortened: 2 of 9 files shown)


MilagrosMarin (approved):
LGTM

PR #234, opened 2026-07-21
docs: diagram operations — fold cascade/trace/restrict derivation into the specs (WIP)

dimitri-yatsenko:
**Status: WIP / draft.**

Reorganizes the diagram-operations documentation on one principle: low-level technical definitions belong in the **specs** (with implementation re-derived from the spec), and **explanation** stays conceptual.

## Changes
- **`reference/specs/master-part.md`** — sharpen the transitive-dependency definition (a foreign key to a master is, in effect, a dependency on the whole composite — master + all parts; downstream tables reference the master alone). Express the Delete Operations rules through the cascade engine (`Diagram.cascade`: master→part is an ordinary downstream edge; `part_integrity="cascade"` = upward U1–U3 walk + materialize + forward re-cascade), cross-linking `cascade.md`. Fix §7.1/§4.4 bold-label list rendering.
- **`reference/specs/diagram.md`** — new **Design Rationale** section: the first-principles derivation that `cascade`/`trace`/`restrict` are one propagation engine (OR-vs-AND, part→master walk only when mutating, materialization), plus **Conformance and open items** (#1496 / #1501 / #1481). Moved here from the explanation doc so the design lives with the normative spec.
- **`explanation/diagram-operations.md`** — reduced to a short conceptual pointer that redirects to the Diagram spec's Design Rationale and the cascade/trace/master-part specs.

## Still to do (WIP)
- `mkdocs.yaml` nav entry for the reduced explanation pointer is **not** in this PR yet.
- `cascade.md` still describes the dependency graph as `networkx.DiGraph`; should read `nx.MultiDiGraph` (2.4 line) to match the derivation.
- Final review of the open-item framing (#1496 / #1501 / #1481).

### DIFF
--- a/src/explanation/diagram-operations.md
+++ b/src/explanation/diagram-operations.md
@@ -1,0 +1,33 @@
+# Diagram Operations: Cascade, Trace, and Restrict
+
+`dj.Diagram` exposes three operations that traverse the dependency graph —
+`cascade`, `trace`, and `restrict` (with `self.upstream` inside `make()` being a
+form of `trace`). They are usually met one at a time, in separate contexts —
+delete impact, lineage, export — so they can look like three unrelated features.
+
+They are not. All three are **one propagation engine aimed in different
+directions**. Their rules — which convergence each uses (`cascade`/`trace` are
+OR closures; `restrict` is an AND selection), when the engine walks up from a
+part to its master (only when the operation *mutates*), and when a restriction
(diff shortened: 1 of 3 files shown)


dimitri-yatsenko:
Superseded by #249, which folds the diagram-traversal derivation into diagram.md around the expand/restrict model (this draft predates that framing). Its in-progress master-part sharpening will be picked up in the follow-up cascade/trace reshape.

PR #238, opened 2026-07-23
docs: state platform support (Linux/macOS/Windows)

dimitri-yatsenko:
> ⚠️ **Do not merge until datajoint-python #1521 and #1522 land.** This documents Windows as a CI-backed supported platform; the claim about the Windows unit-test leg is only accurate once #1522 (the `unit-tests-windows` job) merges, and the underlying file-path fix is #1521.

Records the decision to keep **Windows a supported platform** for DataJoint — including file-protocol object stores and garbage collection — backed by CI, rather than documenting it as unsupported. Consistent with the existing "supported == what we test in CI" framing (the DB version-support policy in `database-backends.md`, from #1497).

## Change

Adds a **Platform support** section to `how-to/installation.md` (plus a Requirements line), stating what CI validates on each OS:
- Full suite incl. DB integration on Linux (Python 3.10 & 3.14).
- Unit suite additionally on Windows (Python 3.10 & 3.14) — covering file-protocol path handling and garbage collection (the class of bug in datajoint-python #1520).
- macOS as a supported development platform.

Wording is deliberately precise about *what runs where* so it doesn't overclaim (Windows runs the unit suite, not the DB-backed integration tests).

## Related
- datajoint-python #1520 (bug), #1521 (fix), #1522 (Windows CI leg).

### DIFF
--- a/src/how-to/installation.md
+++ b/src/how-to/installation.md
@@ -93,6 +93,15 @@ See [Configure Database Connection](configure-database.md) for connection setup.
 - Python 3.10+
 - MySQL 8.0.13+ or PostgreSQL 15+
 - Network access to database server
+- Linux, macOS, or Windows (see [Platform support](#platform-support))
+
+## Platform support
+
+DataJoint runs on **Linux, macOS, and Windows**.
+
+- CI validates the **full test suite — including database integration** — on Linux, across both ends of the supported Python range (3.10 and 3.14).
+- The **unit-test suite additionally runs on Windows** (Python 3.10 and 3.14). This covers OS-specific behavior — notably filesystem path handling for file-protocol object stores and garbage collection, where native path separators must not leak into store paths.
+- macOS is a supported development platform.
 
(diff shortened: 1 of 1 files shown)


MilagrosMarin (approved):
LGTM — verified the three platform claims against CI config (Linux full suite via `test.yaml:29` ubuntu-latest × py310/py314 × mysql 8.0/8.4; Windows unit suite matches #1522's `unit-tests-windows` job exactly; macOS "development platform" correctly reflects `pixi platforms` including `osx-arm64` with no CI leg). Merge gate confirmed still active — #1520 closed, #1521/#1522 both open.

One nit: `## Platform support` is sentence case, while every other `##` in `installation.md` is Title Case (`## Install DataJoint 2.0+`, `## Verify Installation`, `## Database Server`, `## Requirements`, `## Troubleshooting`). Anchor resolves either way; just visible in the rendered TOC next to its siblings.

PR #239, opened 2026-07-23
Define the term as 'schema dimension' (shortened to 'dimension')

dimitri-yatsenko:
Promotes **schema dimension** to the *defined* term, then shortens to *dimension* in-context.

## Why
`dimension` is overloaded — data-warehouse *dimension tables*, OLAP cubes, and *array/tensor dimensions* (the type-system and npy pages already use "dimension" for array axes). Qualifying it as **schema dimension** on definition disambiguates cleanly. This isn't a new coinage: the Entity Integrity page **already** titles the section `## Schema dimensions` and says "lineage through schema dimensions" — this just makes the *defined term* consistent with that.

## Changes (definitional homes only)
- **Entity Integrity → §Schema dimensions:** `A **dimension** is…` → `A **schema dimension** — often shortened to *dimension* — is…`
- **Read Diagrams → §Dimensions and Underlined Names:** the same promotion on the definition sentence.
- Everything else keeps the short form **dimension** (define once, shorten in-context) — so **all headings and anchors are unchanged**.

## Notes
- Overlaps only lightly with the WIP diagram PR #234 (scoped to diagram *operations* — cascade/trace); different sections, low conflict.
- `llms-full.txt` regenerates from source (not hand-edited here).
- The Read Diagrams cell `source` was normalized to a single JSON string by the notebook editor (valid nbformat, identical rendering).

### DIFF
--- a/src/explanation/entity-integrity.md
+++ b/src/explanation/entity-integrity.md
@@ -213,7 +213,8 @@ in a single statement.
 
 ## Schema dimensions
 
-A **dimension** is an independent axis of variation in the data. The rule:
+A **schema dimension** — often shortened to *dimension* — is an independent axis of
+variation in the data. The rule:
 
 > **Any table that introduces a new primary-key attribute introduces a new
 > dimension.**
--- a/src/how-to/read-diagrams.ipynb
+++ b/src/how-to/read-diagrams.ipynb
@@ -389,20 +389,7 @@
    "cell_type": "markdown",
    "id": "cell-dim-md",
    "metadata": {},
-   "source": [
-    "## Dimensions and Underlined Names\n",
-    "\n",
-    "A **dimension** is a new entity type introduced by a table that defines new primary key attributes. Each underlined table introduces exactly **one** dimension—even if it has multiple new PK attributes, together they identify one new entity type.\n",
-    "\n",
-    "| Visual | Meaning |\n",
-    "|--------|--------|\n",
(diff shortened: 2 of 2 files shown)


MilagrosMarin (approved):
The `dimension` → `schema dimension` promotion is well-motivated (the collision with warehouse dimension tables and array axes is real, and the Entity Integrity page already titled the section `## Schema dimensions`), and the "define once, shorten in-context" approach keeps all headings and anchors stable. CI green.

Three mechanical notes on the notebook, none about the wording itself:

1. **Trailing newline dropped** — the diff ends with `\ No newline at end of file`. Every other notebook in the repo (`model-relationships`, `02-schema-design`, `master-part`, and `read-diagrams` on `main`) ends with one, so this is the only file without.

2. **The cell-source normalization is worth undoing.** You flag it as benign nbformat, and it is — but on `main` all 54 cells in `read-diagrams.ipynb` use list-of-strings `source`; on this branch that cell is now the *only* plain-string one. The practical cost is that any future edit to that Markdown cell diffs as a single unsplittable line rather than per-line — which matters for a cell likely to be touched again.

3. **Parenthetical style differs between the two definitional homes** — `entity-integrity.md` uses em-dashes (`— often shortened to *dimension* —`), the notebook uses parens (`(often shortened to *dimension*)`). Same meaning, just reads as drift given both land in this PR.

Approving — none of these are correctness issues.

PR #249, opened 2026-08-10
docs(specs): diagram-traversal algebra derivation (expand/restrict) in diagram.md

dimitri-yatsenko:
Consolidates the full diagram-traversal derivation into the **datajoint-docs specs**, where the normative specs live — relocating it from `datajoint-python`'s `DESIGN-expand-restrict.md` (PR #1524, now closed).

## What

Adds a **"Traversal algebra"** design rationale to `reference/specs/diagram.md` deriving the traversal operations from first principles:

- **one primitive** — propagating a restriction across a foreign key is itself a restriction (`&`), either direction;
- **R1 (edge rule)** — restrict the neighbor by the projected/renamed restricted table; identity fast-path when the FK is the whole PK, unrenamed;
- **R2 (group rule)** — a restriction touching any part lifts to the master and re-expands to all its parts;
- **two operations** — additive `expand(direction)` (of which `cascade` = down, `trace` = up) and subtractive `restrict`;
- **renamed FKs** — the three seed-restriction kinds (materialized / subquery / string) and the relabel fast-path;
- **master↔part** — R2 via relabel-drop + re-expansion, and why a key reaching a part via its master carries no part-specific constraint.

`cascade.md` and `trace.md` now point to this rationale as the downstream/upstream cases, keeping only their operation-specific content.

## Notes
- Single source of truth: the derivation is in `diagram.md`; the other specs reference it rather than re-deriving.
- Terminology/notation follow the house style (restriction/anti-restriction, DataJoint operators, "diagram" not ERD).
- **Supersedes** the old-framing draft #234 (which predates the expand/restrict model).
- Further de-duplication of `cascade.md`/`trace.md`'s older per-rule enumerations (F1–3/U1–3) and the `DiGraph`→`MultiDiGraph` wording is a follow-up reshape.


dimitri-yatsenko:
Revised the derivation to **expand-only**: dropped `restrict` and the additive/subtractive framing. The model is now one operation — `expand(seed, direction in {up,down,both})` (cascade = down, trace = up, export = both) — over R1 (edge) + R2 (group). Multi-condition filtering reduces to seeding a common descendant or a union of expands, so it's a UI-layer composition, not a core primitive. (Note: the Operational Methods section still documents the shipped `Diagram.restrict()`; it's subsumed by `expand` in the model and would be removed when the redesign lands.)

MilagrosMarin (approved):

MilagrosMarin (approved):

PR #260, opened 2026-08-14
docs(data-integrity): domain integrity extends to OAS + codec types

dimitri-yatsenko:
Expands the **Domain integrity** section of [Data Integrity](https://docs.datajoint.com/explanation/data-integrity/) to cover what was only gestured at ("custom codecs attach domain rules … as well").

The added paragraph explains that Object-Augmented Schemas + the codec system carry domain integrity **beyond scalar SQL types**:
- A **codec defines a domain-specific datatype** (a movie, a sorting result, a fitted model) whose contents live in object storage but are addressed and governed by the schema.
- It can run **arbitrary validation checks** on encode (rejecting out-of-domain values, as `enum` does for a category) and expose **typed access methods** on decode.
- Object stores are **not transactional participants**: the codec validates and encodes *before* anything is written, the **row commit is the atomic act** that makes an object referenced, and a failed check aborts the insert with nothing written; unreferenced objects are removed by garbage collection.

Net: domain integrity holds uniformly from `int32`/`enum` to arbitrary structured and object-backed types, under one consistency guarantee. Links to the OAS explanation, the custom-codec how-to, and the codec API spec.

### DIFF
--- a/src/explanation/data-integrity.md
+++ b/src/explanation/data-integrity.md
@@ -75,9 +75,24 @@ each column: `int32`, `int64`, `float32`, `float64`, `decimal(p, s)`,
 `varchar(N)`, `char(N)`, `date`, `datetime`, `uuid`, and `enum(...)` for
 closed categorical sets. Choosing `sex : enum('M', 'F', 'U')` instead of a
 free-text field means an invalid category can never be stored, and
-`decimal(5, 2)` fixes both magnitude and precision. The type system is
-extensible: custom codecs attach domain rules to structured and object-backed
-attributes as well.
+`decimal(5, 2)` fixes both magnitude and precision.
+
+The type system is **extensible**, and this is where DataJoint carries domain
+integrity beyond the scalar SQL types. Through
+[Object-Augmented Schemas](data-pipelines.md#object-augmented-schemas) and the
+[codec system](../how-to/create-custom-codec.md), an attribute can hold a
+domain-specific datatype — a calcium-imaging movie, a spike-sorting result, a
+fitted model — whose contents live in object storage but are addressed and
+governed by the schema. A [codec](../reference/specs/codec-api.md) *defines* that
+datatype: it can run **arbitrary validation checks** when a value is encoded —
+rejecting anything outside the domain's valid set, exactly as `enum` does for a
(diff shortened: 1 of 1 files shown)


MilagrosMarin (dismissed):

MilagrosMarin (dismissed):
Verified the substance: `validate()` runs at `table.py:1369`, before `encode()` at `:1395`, so "the check runs before anything is written" holds. Link targets all resolve and the `#object-augmented-schemas` anchor exists.

Two things:

1. **The PR description still has the pre-`30eb545` claim.** It says "the object write is bound to the same database transaction as its row … row and object commit together or not at all" — which is what that commit was written to correct, and the file now correctly says object stores are not transactional participants. Worth updating the description so a squash-merge message doesn't carry the inaccurate version.

2. **`[garbage collection](referential-integrity.md)` points at a page that doesn't cover it.** `referential-integrity.md` has no mention of garbage collection (its two "orphan" hits are about orphaned records / dangling FKs — a different concept, which makes the mislabel easier to trip over). `../how-to/garbage-collection.md` looks like the intended target. Worth noting §4 of this same file already links that page as "[Referential Integrity](referential-integrity.md)" — text matching target — so it's just this new paragraph's label/target pairing.

dimitri-yatsenko:
Both items addressed: the garbage-collection link now points at `../how-to/garbage-collection.md` (the referential-integrity mislabel is gone), and the PR description's transactional claim is rewritten to match what `30eb545` corrected in the file — object stores are not transactional participants; the row commit is the atomic act — so a squash message carries the accurate version. One-line commit; re-review requested since the push resets the approval.

MilagrosMarin (approved):
Both verified. The link now points at `../how-to/garbage-collection.md` — target resolves, and you correctly left §4's existing `[Referential Integrity](referential-integrity.md)` alone since that pairing was already text-matching-target. Description reads accurately now, and "a failed check aborts the insert with nothing written" is exactly right given `validate()` precedes `encode()`.

One micro-nit, not worth a push on its own: "validates and encodes before anything is written" — encoding *is* the object write, so strictly it's validate → encode (object written) → row commit. The clauses after it make the model clear, so it reads fine in context.

Re-approving.

